PR #4, opened 2022-01-16
Add max data rate option for when idle.

PhenomRetroShare:
With [RetroShare#994](https://github.com/RetroShare/RetroShare/pull/994)

### DIFF
--- a/src/retroshare/rsconfig.h
+++ b/src/retroshare/rsconfig.h
@@ -382,16 +382,16 @@ public:
 	 * @param[in] opMode new opearting mode
 	 * @return
 	 */
-	virtual bool     setOperatingMode(RsOpMode opMode) = 0;
+	virtual bool setOperatingMode(RsOpMode opMode) = 0;
 
 	/**
 	 * @brief setOperatingMode set the current operating mode from string
 	 * @param[in] opModeStr new operating mode as string
 	 * @return
 	 */
-    virtual bool     setOperatingMode(const std::string &opModeStr) = 0;
+	virtual bool setOperatingMode(const std::string &opModeStr) = 0;
 
-    /* Data Rate Control - to be moved here */
+	/* Data Rate Control */
 	/**
 	 * @brief SetMaxDataRates set maximum upload and download rates
 	 * @jsonapi{development}
@@ -399,7 +399,20 @@ public:
 	 * @param[in] upKb upload rate in kB
 	 * @return returns 1 on succes and 0 otherwise
 	 */
-    virtual int SetMaxDataRates( int downKb, int upKb ) = 0;
+	RS_DEPRECATED_FOR("setMaxDataRates( int inKb, int outKb, int inKbWhenIdle, int outKbWhenIdle)") \
+	virtual int SetMaxDataRates( int downKb, int upKb) = 0;
+
+	/* Data Rate Control */
+	/**
+	 * @brief setMaxDataRates set maximum upload and download rates
+	 * @jsonapi{development}
+	 * @param[in] inKb download rate in kB
+	 * @param[in] outKb upload rate in kB
+	 * @param[in] inKbWhenIdle download rate in kB When Idle
+	 * @param[in] outKbWhenIdle upload rate in kB When Idle
+	 * @return returns 1 on succes and 0 otherwise
+	 */
+	virtual int setMaxDataRates( int inKb, int outKb, int inKbWhenIdle, int outKbWhenIdle) = 0;
 
 	/**
 	 * @brief GetMaxDataRates get maximum upload and download rates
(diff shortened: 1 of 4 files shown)


src/retroshare/rsconfig.h
```diff
@@ -395,20 +395,24 @@ class RsServerConfig
...
-	 * @param[in] downKb download rate in kB
-	 * @param[in] upKb upload rate in kB
+	 * @param[in] inKb download rate in kB
+	 * @param[in] outKb upload rate in kB
+	 * @param[in] inKbWhenIdle download rate in kB When Idle
+	 * @param[in] outKbWhenIdle upload rate in kB When Idle
 	 * @return returns 1 on succes and 0 otherwise
 	 */
-    virtual int SetMaxDataRates( int downKb, int upKb ) = 0;
+    virtual int SetMaxDataRates( int inKb, int outKb, int inKbWhenIdle, int outKbWhenIdle ) = 0;
```
G10h4ck, line 404: This will break API retro-compatibility, what's worse is that JSON API user will not even notice, they will just get random data rates set.

In this case I see two possibilities, 
1)  Provide a new method
`virtual int setMaxDataRates( int inKb, int outKb, int inKbWhenIdle, int outKbWhenIdle ) = 0;`
In this case as the former is not coherent with naming convention you can even avoid the V2 at the end of the name thanks to that sS (we cannot expose method with same name in JSON API) and mark the former `SetMaxDataRates` as deprecated

2) Provide a method to just set the idle rate and leave `SetMaxDataRates` as is without even touching parameters names

AFAIU 1 is what make more sense but i am not 100% sure

src/retroshare/rsconfig.h
```diff
@@ -395,20 +395,24 @@ class RsServerConfig
...
 	 * @brief GetMaxDataRates get maximum upload and download rates
 	 * @jsonapi{development}
 	 * @param[out] inKb download rate in kB
 	 * @param[out] outKb upload rate in kB
+	 * @param[out] inKbWhenIdle download rate in kB When Idle
+	 * @param[out] outKbWhenIdle upload rate in kB When Idle
 	 * @return returns 1 on succes and 0 otherwise
 	 */
-    virtual int GetMaxDataRates( int &inKb, int &outKb ) = 0;
+    virtual int GetMaxDataRates( int &inKb, int &outKb, int &inKbWhenIdle, int &outKbWhenIdle ) = 0;
```
G10h4ck, line 415: Same things I said for `SetMaxDataRates` apply here too

G10h4ck (changes requested):
The important part are the inline comments, in addition to those I am not convinced about the matching between "temporarily reducing bandwidth usage" and the wording "idle" maybe we should think a better wording.

src/retroshare/rsconfig.h
```diff
@@ -389,17 +389,30 @@ class RsServerConfig
...
 	 * @jsonapi{development}
-	 * @param[in] downKb download rate in kB
-	 * @param[in] upKb upload rate in kB
+	 * @param[in] inKb download rate in kB
+	 * @param[in] outKb upload rate in kB
 	 * @return returns 1 on succes and 0 otherwise
 	 */
-    virtual int SetMaxDataRates( int downKb, int upKb ) = 0;
+	RS_DEPRECATED_FOR("setMaxDataRates( int inKb, int outKb, int inKbWhenIdle, int outKbWhenIdle)") \
+	virtual int SetMaxDataRates( int inKb, int outKb) = 0;
```
G10h4ck, line 403: Param names must remain the same too, otherwise JSON API retro-compatibility will be broken

src/retroshare/rsconfig.h
```diff
@@ -389,17 +389,30 @@ class RsServerConfig
 	 * @param[in] opModeStr new operating mode as string
 	 * @return
 	 */
-    virtual bool     setOperatingMode(const std::string &opModeStr) = 0;
+	virtual bool     setOperatingMode(const std::string &opModeStr) = 0;
```
G10h4ck, line 392: now that you touched it can you fix the indentation and spacing too? Just one tab before `virtual` just one space between `bool` and `setOperatingMode`

src/rsserver/p3serverconfig.cc
```diff
@@ -28,70 +28,73 @@
...
 	rsConfig = this;
 }
 
 
-p3ServerConfig::~p3ServerConfig() 
-{ 
-	return; 
+p3ServerConfig::~p3ServerConfig()
+{
+	return;
```
G10h4ck, line 55: change this to `p3ServerConfig::~p3ServerConfig() = default;`

src/rsserver/p3serverconfig.cc
```diff
@@ -28,70 +28,73 @@
...
+	}
 
 	/* enable operating mode */
 	RsOpMode opMode = getOperatingMode();
 	switchToOperatingMode(opMode);
 }
 
 
-#define RS_CONFIG_ADVANCED_STRING 	"AdvMode"
+#define RS_CONFIG_ADVANCED_STRING   "AdvMode"
```
G10h4ck, line 97: constexpress here too

G10h4ck (changes requested):
Just a few more minor fixes ad it's ok for me, read inline comments for details

src/rsserver/p3serverconfig.cc
```diff
@@ -28,70 +28,73 @@
...
-
 	RsStackMutex stack(configMtx); /******* LOCKED MUTEX *****/
-
-	mUserLevel = RsConfigUserLvl::NEW; /* START LEVEL */
-	mRateDownload =  DEFAULT_DOWNLOAD_KB_RATE;
-	mRateUpload = DEFAULT_UPLOAD_KB_RATE;
-
-	mOpMode = RsOpMode::FULL;
-
 	rsConfig = this;
```
csoler, line 49: This is really terrible coding IMO (I mean initializing a global variable from within the constructor of one instance of that global variable). If you have 2 mins, move that to rsInit.cc where all rsSomthing objects are initialized.

src/retroshare/rsconfig.h
PhenomRetroShare: Done

PhenomRetroShare:
Updated

G10h4ck (approved):

G10h4ck:
Thanks for the contribution @PhenomRetroShare

PR #13, opened 2022-02-15
Distribute libretroshare as an Android library AAR

G10h4ck:
We have been capable of compiling libretroshare for Android since a while, but it wasn't easy for other developer to bundle it in their own applications.
Compiling libretroshare for Android require a complex toolchain that is not easy to reproduce, and integrate with third party build systems.
Android documentation says to distribute libraries as AAR, which is similar to Android APK packages but tailored to libraries distribution.
This Pull request improve the libretroshare build system so it compile an AAR, and provide tools for a more reproducible way of doing it, plus add capability to deploy the produced AARs directly to a maven repository hosted by gitlab.com, so now it becames really easy and stright forward for Android applications to include libretroshare as a dependency, just by declaring in in their build system such as gradle or maven.


csoler:
Seem ok to me, although I'm certainly no expert here.

PR #14, opened 2022-02-18
Improve gradle build and fix crash in rsGlobalShutDown()

G10h4ck:


src/rsserver/p3face-config.cc
```diff
@@ -81,42 +81,31 @@ void RsServer::startServiceThread(RsTickingThread *t, const std::string &threadN
...
 	coreReady = false;
-	// TODO: cache should also clean up old files
 
-	ConfigFinalSave(); // save configuration before exit
-
-	mPluginsManager->stopPlugins(pqih);
+	if(wasReady)
+	{
+		// save configuration before exit
+		if(wasReady) ConfigFinalSave();
```
csoler, line 90: why re-test this?

PR #27, opened 2022-05-10
JSON API avoid parsing unuseful dirs

G10h4ck:


csoler:
Seems ok. Unless you want to add anything, it can be merged.

G10h4ck:
Thanks for reviewing, at some point we may remove also the `unused` and `unfinished` directories but let's live that for another PR in the future

PR #34, opened 2022-06-22
Fix + SAMv3 update

sehraf:
- There was an leftover `exit -1;` which crashes RS. Remove it since RS handles the error case well.
- User "I2P" pointed out that there are connection problems involving i2p encrypted lease sets. [i2psam](https://github.com/i2p/i2psam/blob/a09ab6b4841031b18be88b9b32d89d6be881daef/i2psam.h#L54) sets this per default `"i2cp.leaseSetEncType=0,4"` , do the same in RS
- A warning is silenced when shutting down. The warning is _expected_ when shutting down and therefore missleading.


csoler:
seems all fine to me. Should we merge?

sehraf:
please do so :+1:

PR #53, opened 2022-10-29
fixed bug causing WebUI to not properly start when -U [user id] is used in the commandline.

csoler:


G10h4ck:
`mRestartReqTS` is declared to be an `std:atomic` https://github.com/RetroShare/libretroshare/blob/master/src/jsonapi/jsonapi.h#L219 , now that a mutex is used it don't need to be that anymore, a plain `rstime_t` will suffice. A part from that seems ok to me thans also for commenting the code ;)

Cheers

G10h4ck:
now that you are touching it please remove also de duplicated include here
https://github.com/RetroShare/libretroshare/blob/master/src/jsonapi/jsonapi.h#L40

PR #54, opened 2022-11-02
WIP: Fix crash in RsBrodacastDiscovery

G10h4ck:
Deal properly with corrupted broadcast discovery packets received from the
  network instead of crashing
Improve error reporting and documentation of rs_malloc

Should fix error reported by elRepo.io developers
https://gitlab.com/elRepo.io/elRepo.io-android/-/issues/62

### DIFF
--- a/src/services/broadcastdiscoveryservice.cc
+++ b/src/services/broadcastdiscoveryservice.cc
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * RetroShare Broadcast Domain Discovery                                       *
  *                                                                             *
- * Copyright (C) 2019-2021  Gioacchino Mazzurco <gio@altermundi.net>           *
- * Copyright (C) 2019-2021  Asociación Civil Altermundi <info@altermundi.net>  *
+ * Copyright (C) 2019-2022  Gioacchino Mazzurco <gio@altermundi.net>           *
+ * Copyright (C) 2019-2022  Asociación Civil Altermundi <info@altermundi.net>  *
  *                                                                             *
  * This program is free software: you can redistribute it and/or modify        *
  * it under the terms of the GNU Lesser General Public License as              *
@@ -25,6 +25,7 @@
 #include <chrono>
 #include <vector>
 #include <iostream>
+#include <cstdlib>
 
 #include "services/broadcastdiscoveryservice.h"
 #include "retroshare/rspeers.h"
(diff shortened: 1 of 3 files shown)


src/services/broadcastdiscoveryservice.cc
```diff
@@ -67,14 +68,46 @@ struct BroadcastDiscoveryPack : RsSerializable
...
+		auto bdp = std::make_unique<BroadcastDiscoveryPack>();
+		bdp->serial_process(RsGenericSerializer::DESERIALIZE, ctx);
+		if(ctx.mOk) return bdp;
+
+		if(!ec)
+		{
+			RS_FATAL( "Attempteted from invalid string ",
+			          std::errc::invalid_argument );
+			print_stacktrace();
+			exit(static_cast<int>(std::errc::invalid_argument));
```
csoler, line 106: same here.

src/services/broadcastdiscoveryservice.h
```diff
@@ -76,8 +77,18 @@ class BroadcastDiscoveryService :
...
+	 * @param[in] ipp peer IP and port
+	 * @param[in] uData serialized data associated to the peer
+	 * @param[out] ec Optional storage for eventual error code,
+	 *	meaningful only on failure, if a nullptr is passed ther error is treated
+	 *	as fatal downstream, otherwise it bubble up to be treated upstream
+	 * @return nullptr on failure, pointer to the generated result otherwise
+	 */
+	static std::unique_ptr<RsBroadcastDiscoveryResult> createResult(
+	        const UDC::IpPort& ipp, const std::string& uData,
+	        rs_view_ptr<std::error_condition> ec = nullptr );
```
csoler, line 91: That looks like a complicated construct for something that could simply be std::error_condition*

src/util/rsmemory.h
```diff
@@ -146,6 +148,7 @@ template<typename T = void> rs_owner_ptr<T> rs_malloc(
 		{
 			RS_ERR( "A chunk of size larger than ", SAFE_MEMALLOC_THRESHOLD,
 			        " was requested" );
+			print_stacktrace();
 			exit(static_cast<int>(std::errc::argument_out_of_domain));
```
csoler, line 152: The exit here is justified (contrary to the other ones), because allocating a large memory block is an error in the code. We could of course return null and let the caller take care of it.

src/services/broadcastdiscoveryservice.h
G10h4ck: it is a decorated way of writing `std::error_condition*` the decoration is just to informa the pointer lifetime is handled by the caller but it is just a pointer

csoler: ah yes. I always forget that ;-)

src/services/broadcastdiscoveryservice.cc
```diff
@@ -67,14 +68,46 @@ struct BroadcastDiscoveryPack : RsSerializable
...
+	        const std::string& st,
+	        rs_view_ptr<std::error_condition> ec )
 	{
+		if(st.empty())
+		{
+			if(!ec)
+			{
+				RS_FATAL("Attempteted from empty string ", std::errc::no_message);
+				print_stacktrace();
+				exit(static_cast<int>(std::errc::no_message));
```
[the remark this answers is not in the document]
G10h4ck: that case happens only when ask to the function to parse something we generated, that should never fail, instead when the function is called to parse stuff from the network the error_condition pointer is not null so the error bubble up without crash, and the error is handled gracefully

src/services/broadcastdiscoveryservice.cc
G10h4ck: same here

PR #58, opened 2022-11-09
Bug fixing

csoler:


src/pgp/pgphandler.cc
```diff
@@ -59,6 +59,9 @@ void PGPHandler::setPassphraseCallback(PassphraseCallback cb)
 PGPHandler::PGPHandler(const std::string& pubring, const std::string& secring,const std::string& trustdb,const std::string& pgp_lock_filename)
 	: pgphandlerMtx(std::string("PGPHandler")), _pubring_path(pubring),_secring_path(secring),_trustdb_path(trustdb),_pgp_lock_filename(pgp_lock_filename)
 {
+    _trustdb_changed = false;
+    _pubring_changed = false;
+    _pubring_last_update_time = time(NULL) ;
```
G10h4ck, line 64: shouldn't those go into the initializer list (just before `{` ) instead of the constructor body?

csoler: done

PR #69, opened 2023-01-26
experimental hash system to speedup when full rehashing is forced

csoler:


csoler:
Not the best way to handle it.

PR #75, opened 2023-02-01
Delays turtle forward probabilities calculation until needed

jolavillette:
80 pct of received TRs are bouncing and thus dropped.
Those TRs don't need forward probability calculation.
With this PR the calculation is done only when the TR is new.

### DIFF
--- a/src/turtle/p3turtle.cc
+++ b/src/turtle/p3turtle.cc
@@ -1606,34 +1606,6 @@ void p3turtle::handleTunnelRequest(RsTurtleOpenTunnelItem *item)
 	if(TS_request_bounces.find(item->request_id) != TS_request_bounces.end())
 		TS_request_bounces[item->request_id].push_back(time(NULL)) ;
 #endif
-	// TR forwarding. We must pay attention not to flood the network. The policy is to force a statistical behavior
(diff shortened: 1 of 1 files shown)


src/turtle/p3turtle.cc
```diff
@@ -1762,6 +1734,35 @@ void p3turtle::handleTunnelRequest(RsTurtleOpenTunnelItem *item)
...
+	// 	forward_probability	: final probability of forwarding the packet, per peer.
+	//
+	// When the number of peers increases, the speed limit is reached faster, but the behavior per peer is the same.
+	//
+
+	float forward_probability ;
+
+	{
+		RsStackMutex stack(mTurtleMtx); /********** STACK LOCKED MTX ******/
+		_traffic_info_buffer.tr_dn_Bps += RsTurtleSerialiser().size(item);
```
csoler, line 1756: This needs to be kept before exit (same place than before this PR).

jolavillette: You are right. I will update the PR.

jolavillette: I wrongly introduced authssl.cc fix in the PR. I don't know how to revert that.

PR #79, opened 2023-02-16
Improvement of GxsId service

csoler:
added reject list in p3idservice.cc to put GxsIds and updates that are faulty, so as to not re-ask them to friends


csoler:
This has already been tested in PR 77

PR #88, opened 2023-05-04
Make turtle search results be sent as events, so that the webUI can catch them.

csoler:


zelfroster:
Great, I will start working on its implementation in webui.

zelfroster:
I can't believe I am so dumb. Guess, I have a lot to learn.
I was trying to call turtleSearch from so much time and wondering why is it not working, trying so many ways. And then when I came here to check the code changes then I realised that I haven't pulled the changes yet and built rs correctly, smh :(

PR #107, opened 2023-07-04
file upload stats (lib)

RetroPooh:


src/ft/ftserver.cc
```diff
@@ -2385,3 +2394,24 @@ std::error_condition ftServer::parseFilesLink(
 	if(tft) collection = *tft;
 	return ec;
 }
+
+uint64_t ftServer::getCumulativeUpload(RsFileHash hash)
+{
+	std::map<RsFileHash,uint64_t>::iterator it = cumulative_uploaded.find(hash) ;
```
csoler, line 2400: looks like a mutex is missing here.

src/ft/ftserver.cc
```diff
@@ -2385,3 +2394,24 @@ std::error_condition ftServer::parseFilesLink(
...
+	if(it != cumulative_uploaded.end())
+		return it->second;
+	return 0;
+}
+
+uint64_t ftServer::getCumulativeUploadAll()
+{
+	uint64_t all = 0;
```
csoler, line 2408: same here

csoler:
Make sure to always use a mutex protection when reading/writing to a std::map. Otherwise it's bound to crash when you access from different threads.

Another problem is that you're not saving the upload stats. So they are lost accross restarts, which somewhat kills the entire idea since upload stats are supposed to be long-term stats. But ftServer does not have a saving system.

All accounted for, it seems that it would make much more sense to store them with p3FileDatabase in ft_database.cfg. 
That means:
- add to p3FileDatabase class the functions to update the upload stats, as called from ftServer (no mutex needed in ftServer for that since the one is p3FileDatabase will be used)
- create a new RsItem in rsfilelistitems.h to store the upload stats.
- add the proper lines to load/save these stats in p3FileDatabase::loadList() and saveList() using these created items

defnax:
yes would be nice save the stats and reload after restart

defnax:
@RetroPooh you will update this pr?

defnax:
@RetroPooh

defnax:
@RetroPooh

defnax:
when we get this stats in retroshare? we need this

defnax:
@csoler i have added before the mutex this is right or

image

csoler:
yes this is fine.

csoler:
I like the idea of keeping these stats. Last time I looked at it, it still needed saving that data. Of course we need to only save the stats for files that actually have been uploaded (in order to limit the file size) and using a std::map for that is indeed a good idea.

defnax:
Hi, you will reimplement this? else i can make a new pr with the mutex

defnax:
new pr for this : https://github.com/RetroShare/libretroshare/pull/142

defnax:
Done in #220 ✅
Close @csoler

PR #109, opened 2023-07-04
router statistics - filter out gxs tunnels from file tunnels

RetroPooh:


defnax:
@csoler

src/gxs/rsgxsnettunnel.cc
```diff
@@ -1467,6 +1467,10 @@ void RsGxsNetTunnelService::getStatistics(
     bias = mRandomBias ;
 }
 
+bool RsGxsNetTunnelService::isGXSHash(RsFileHash hash) const
+{
+	return mHandledHashes.find(hash) != mHandledHashes.end();
```
csoler, line 1472: A mutex is missing here as you access a std::map. But having a mutex here would likely cause a deadlock randomly since you call this function from p3turtle and p3turtle is always locked after GXS tunnel and not the reverse.

defnax:
@RetroPooh no fix?

PR #128, opened 2023-11-30
Allow to override openssl security level

hunbernd:


csoler:
Thx for sorting this out. I actually made it simpler in a new PR 141 where RS tries level 0 when it cannot load the certificate with levels 1 and 2, and warn the user. New users will always get a SHA256-signed certificate, so it's not really causing security problems to start a SHA1-signed node.

PR #132, opened 2023-12-30
Fixed crash on on tor profile generation on macOS

defnax:


src/tor/TorManager.cpp
```diff
@@ -933,7 +933,7 @@ void RsTor::setHiddenServiceDirectory(const std::string& dir)
 
 TorManager *RsTor::instance()
```
csoler, line 934: I'm wondering why this #ifdef is needed: if syscall(SYS_gettid) is not available, then we need to find some other way, because calling TorManager::instance() in a thread is not a good idea. If that happens, it means that the code that calls instance() is wrong.

defnax: i dont know but now crash is solved and tor is started on macos,now on macos we can use tor with rs
but seems you added the if defs for windows too?, it was only for linux?

defnax: before i added the macos def i get warning in qt creator
```
TorManager.cpp:937:24: warning: 'syscall' is deprecated: first deprecated in macOS 10.12 - syscall(2) is unsupported; please switch to a supported interface. For SYS_kdebug_trace use kdebug_signpost().
unistd.h:745:6: note: 'syscall' has been explicitly marked deprecated here
```

defnax: https://github.com/google/glog/issues/185

csoler: the warning above is not a crash.
Was the crash due to the assert() call?

defnax: Yes reported by a user in rs network.
Look on the retroshare bugs forum

csoler: well then it means 2 things:
1 - the assert *is* useful because it catches something that would cause deep trouble
2 - there is a bug because instance() shouldn't be called the first time from a thread

PR #140, opened 2024-01-31
fixup! fixed inconsistent code in rstypeserializer.h

G10h4ck:
Fix the fixes ;)


G10h4ck:
now it checks also for duplicated keys in the serialized item

PR #150, opened 2024-09-10
Added Chat description for set chat rules & chat links & infos

defnax:


zapek:
This will break. You can't add new fields to items like that. And adding yet another item type just for that is ridiculous. I suggest to just drop this PR and treat a '\n' in the topic as implying that there's a description?

defnax:
closed

PR #160, opened 2025-01-10
fixed backward compatibility with sha1 certs

csoler:


JohannesKleine:
- I can confirm, that it is now possible to add "old" clients with long certificate format. Short format was already working.
- Tested: same build (this pull request) version can still add each other with short and long certificate versions.
- Profile name changed from Numbers to the Identity.

JohannesKleine:
Observed: two times in Network>Keyring tab while sorting random columns, columns labels are gone and at the same time empty lines (looks like three) are displayed below. For testing purposes both clients run under the same user and share the same .retroshare folder just with different profiles in it. There might be a correlation between the shared keyrings in /rnp in this case, being used by two different instances at the same time. As best of my knowledge the three empty lines where observed before this pull request. I would rule this as highly likely not related to this pull request.

PR #165, opened 2025-02-12
Improve file transfer speed

jolavillette:
Increases the maximum transfer rate to 100 MB/s instead of 10 MB/s in ftcontroller
Improves the transfer rate stability by using millisecond instead of second granularity in rsthreads
Tests show that these changes allow transfer rate of at least 14 MB/s compared to previous 9 MB/s
Please comment

### DIFF
--- a/src/util/rsthreads.cc
+++ b/src/util/rsthreads.cc
@@ -242,7 +242,7 @@ RsQueueThread::RsQueueThread(uint32_t min, uint32_t max, double relaxFactor )
     :mMinSleep(min), mMaxSleep(max), mRelaxFactor(relaxFactor)
 {
     mLastSleep = (uint32_t)mMinSleep ;
-    mLastWork = time(NULL) ;
+    mLastWork = 1000 * getCurrentTS() ;
 }
 
 void RsQueueThread::threadTick()
--- a/src/util/rsthreads.cc
+++ b/src/util/rsthreads.cc
@@ -67,6 +67,21 @@ int RS_pthread_setname_np(pthread_t __target_thread, const char *__buf) {
 	#include <iostream>
 #endif
 
+#include <sys/time.h>
+static double getCurrentTS()
+{
+#ifndef WINDOWS_SYS
+        struct timeval cts_tmp;
+        gettimeofday(&cts_tmp, NULL);
+        double cts =  (cts_tmp.tv_sec) + ((double) cts_tmp.tv_usec) / 1000000.0;
+#else
+        struct _timeb timebuf;
+        _ftime( &timebuf);
+        double cts =  (timebuf.time) + ((double) timebuf.millitm) / 1000.0;
+#endif
+        return cts;
+}
+
 /*static*/ void* RsThread::rsthread_init(void* p)
 {
 	RsThread* thread = reinterpret_cast<RsThread *>(p);
@@ -237,7 +252,7 @@ void RsQueueThread::threadTick()
     {
         doneWork = true;
     }
-    time_t now = time(NULL);
+    uint64_t now = 1000 * getCurrentTS();
     if (doneWork)
     {
         mLastWork = now;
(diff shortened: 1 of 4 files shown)


src/util/rsthreads.cc
```diff
@@ -249,8 +264,8 @@ void RsQueueThread::threadTick()
     }
     else
     {
-        uint32_t deltaT = now - mLastWork;
-        double frac = deltaT / mRelaxFactor;
+        uint64_t deltaT = now - mLastWork;
+        double frac = (deltaT / 1000) / mRelaxFactor;
```
csoler, line 268: you divide an int by an int => is that intentional?

src/util/rsthreads.cc
```diff
@@ -227,7 +242,7 @@ RsQueueThread::RsQueueThread(uint32_t min, uint32_t max, double relaxFactor )
     :mMinSleep(min), mMaxSleep(max), mRelaxFactor(relaxFactor)
 {
     mLastSleep = (uint32_t)mMinSleep ;
-    mLastWork = time(NULL) ;
+    mLastWork = 1000 * getCurrentTS() ;
```
csoler, line 245: it would be helpful to rename mLastWork into mLastWork_ms, so that the units are obvious when reading the code.

src/util/rsthreads.cc
```diff
@@ -249,8 +264,8 @@ void RsQueueThread::threadTick()
     }
     else
     {
-        uint32_t deltaT = now - mLastWork;
-        double frac = deltaT / mRelaxFactor;
+        uint64_t deltaT = now - mLastWork;
```
csoler, line 267: at midnight, this is going to cause an int overflow: mLastWork=86399 (last second of previous day) and now = 0 => deltaT = approximately 10^31

csoler (changes requested):

src/util/rsthreads.cc
jolavillette: actually no, because 1000*getCurrentTS() is the number of ms since 1/1/70, it never returns to 0

csoler: Ah you're right! So getCurrentTS() is a double, which means it has 52 bits of precision. timebuf.time is 64 bits but effectively uses 32, and adding milliseconds you need approximately 42 bits. That gives us some slack. Making getCurrentTS_ms() returning miliseconds as a uint64_t would probably avoid an implicit cast (line 255) and give you more space, but it's not critical here.

src/util/rsthreads.cc
jolavillette: That was not intentional
I should have written frac = deltaT / (1000 * mRelaxFactor) to avoid the int by int division

But actually the whole mRelaxFactor and frac thing is a bad idea.
The original formule was frac = deltaT / mRelaxFactor, where deltaT was the time in seconds elapsed since last data were send to the peer, and mRelaxFactor was always 0.5.
Because of that, even be for similar events deltaT could be either 0 and frac was 0 as the next sleep time was slowly increasing by 5, or 1 and frac was 2 and the next sleep time was 1000 ms. This is plain bad algorythm, with a heavy random threshold effect.
Initially I respected the coder will, because I wanted to change a little code as possible in rsthreads.cc but now I realize how bad this code was.
We need to entirely remove this mRelaxFactor and frac thing and just increase a little bit the sleeping time every time  no work was done with the following formula:
 mLastSleep += (uint32_t) ((mMaxSleep-mMinSleep) * 0.05));
Much more simple, much more efficient, no reandom threshold effect.

src/util/rsthreads.cc
jolavillette: agreed

src/util/rsthreads.cc
jolavillette: I use getCurrentTS() written by DrBob since long. It very well suits my need everytime I need to do calculations. It is a pretty useful method that should  be available everywhee in the code without the need to redefine it.

PR #167, opened 2025-03-29
Added separate notify for distant chat

defnax:


src/chat/distantchat.cc
```diff
@@ -200,6 +201,7 @@ void DistantChatService::notifyTunnelStatus( const RsGxsTunnelId& tunnel_id, uin
         
     case RsGxsTunnelService::RS_GXS_TUNNEL_STATUS_REMOTELY_CLOSED:	RsServer::notify()->notifyChatStatus(ChatId(DistantChatPeerId(tunnel_id)),"Tunnel is down...") ;
         								RsServer::notify()->notifyPeerStatusChanged(tunnel_id.toStdString(),RS_STATUS_OFFLINE) ;
+        								RsServer::notify()->notifyDistantChatStatus(ChatId(DistantChatPeerId(tunnel_id))) ;
```
csoler, line 204: I don't understand the goal of this line: it does exactly what the notifyChatStatus does 2 lines before: same tunnel  id, with message "tunnel is down". So there's no point in having both notify calls.

src/chat/distantchat.cc
```diff
@@ -192,6 +192,7 @@ void DistantChatService::notifyTunnelStatus( const RsGxsTunnelId& tunnel_id, uin
         
     case RsGxsTunnelService::RS_GXS_TUNNEL_STATUS_CAN_TALK:    	RsServer::notify()->notifyChatStatus(ChatId(DistantChatPeerId(tunnel_id)),"Tunnel is secured. You can talk!") ;
         								RsServer::notify()->notifyPeerStatusChanged(tunnel_id.toStdString(),RS_STATUS_ONLINE) ;
+        								RsServer::notify()->notifyDistantChatStatus(ChatId(DistantChatPeerId(tunnel_id))) ;
```
csoler, line 195: same here

G10h4ck:
Also RsNotify is kept just for retrocompatibility, no new notifications should be added to it anymore. If some new notifications are needed/usefult RsEvents must be used.

PR #180, opened 2025-07-11
add missing include for Windows

JohannesKleine:
tested on: Windows 11 with windos-msys2/buil.bat, Debian 12, MacOs 15.5


thunder2:
Already fixed but thanks anyway

PR #192, opened 2025-08-04
CHANGE: deconstructor add ~

JohannesKleine:


csoler:
This change is not valid: RsControl is a static class, meaning that only instance() can be called and the constructor can only be called internally (by instance()  to ensure there's a single element of class RsControl at once). You may have missed that point.

PR #196, opened 2025-09-09
fix botan path

defnax:
with this changes it gets work with botan@2 bigint issue but on local mac machine
but on workflow installed botan include & lib are installed on different path not sure how we can handle this?

image

`/opt/homebrew/Cellar/botan@2/2.19.5 
`

```
For compilers to find botan@2 you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/botan@2/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/botan@2/include"
```


defnax:
@JohannesKleine

PR #197, opened 2025-10-13
fix group flags when changing pseudonymity status of an identity

dbear496:
When the `pseudonimous` argument to `RsIdentity::updateIdentity` differs from the original pseudonymity status of the identity, the old group flags must be discarded to avoid having both `GXS_SERV::FLAG_PRIVACY_PRIVATE` and `GXS_SERV::FLAG_PRIVACY_PUBLIC` flags at the same time.


csoler:
sorry I took so much time to review this.

PR #199, opened 2025-11-22
no V07_NON_BACKWARD_COMPATIBLE_CHANGE_006 in retroshare.pri

Passw:
https://github.com/RetroShare/RetroShare/blob/master/retroshare.pri#L310-L345


csoler:
good catch!

PR #202, opened 2025-11-28
rsdir.cc: remove trailing slash/backslash in directory names to avoid mingw64 _wstat fail

jolavillette:
rsdir.cc: remove trailing slash/backslash in directory names to avoid mingw64 _wstat fail


csoler:
My fault, I didn't see it!  So it's all good as far as I'm concerned.

PR #255, opened 2026-01-28
Wiki moderator management

samuel-asleep:
## Summary

This PR refines wiki moderator management and snapshot retrieval by removing duplicated state, eliminating inefficient scans, and ensuring snapshot content always resolves to the latest edit in an edit hierarchy.

All actionable review feedback from **#250** has been addressed.

---

### Wiki moderator management
- Removed duplicated moderator list state.
- `RsWikiCollection` now uses `mModeratorTerminationDates` as the single source of truth.
- Active moderators are represented by a `0` termination timestamp, eliminating redundant data structures and linear scans.

### Snapshot content retrieval
- Snapshot APIs now require `grpId`, avoiding all-group message scans.
- `getSnapshotContent` and `getSnapshotsContent` fetch only the requested message IDs (meta + data).
- Snapshot resolution follows `mOrigMsgId` and returns the most recent edit by publish timestamp.

---

## Why

- Prevent duplicated moderator state and reduce maintenance complexity.
- Avoid inefficient “get all and filter” message access patterns.
- Ensure snapshot requests reliably return the latest edit in a wiki edit chain.
- Align snapshot APIs with existing group-scoped access patterns used elsewhere in the codebase.

---

## Review feedback addressed (from #250)

- ✅ Removed duplicated moderator list and relied solely on the existing map.
- ✅ Replaced linear list scans with map-based lookups in `p3wiki`.
- ✅ Avoided fetching all groups/messages for snapshot resolution.
- ✅ Added `grpId` to snapshot APIs to scope message access correctly.
- ✅ Resolved snapshot lineage using `mOrigMsgId` and returned the latest edit.

**Note:**  
The concern about direct `getDataStore()` usage in `p3wiki` is acknowledged and partially addressed. Fully removing this dependency would require a broader refactor to token-based request APIs and is intentionally left out to keep this PR focused and reviewable.

---

## How tested

- Code reviewed for correctness and consistency with existing wiki and GXS access patterns.
- Verified no remaining linear scans or duplicated moderator state.
- Confirmed snapshot APIs now operate on scoped message requests only.

---

## Notes

- Public wiki snapshot APIs now include `grpId`.
- No behavioral change for existing consumers beyond improved performance and correctness.

### DIFF
--- a/src/services/p3wiki.cc
+++ b/src/services/p3wiki.cc
@@ -515,58 +515,41 @@ bool p3Wiki::checkModeratorPermission(const RsGxsGroupId& grpId, const RsGxsId& 
 	return isActiveModerator(grpId, authorId, editTime);
 }
 
-bool p3Wiki::getCollectionData(const RsGxsGroupId& grpId, RsWikiCollection& collection) const
+bool p3Wiki::getCollectionData(const RsGxsGroupId& grpId, RsWikiCollection& collection)
 {
-	std::map<RsGxsGroupId, RsNxsGrp*> grpMap;
-	grpMap[grpId] = nullptr;
-	std::map<RsGxsGroupId, RsNxsGrp*>::const_iterator grp_it;
-	
-	if (!getDataStore()->retrieveNxsGrps(grpMap, true) || grpMap.end() == (grp_it = grpMap.find(grpId)) || !grp_it->second)
-		return false;
-	
-	RsNxsGrp* grpData = grp_it->second;
-
-	std::unique_ptr<RsNxsGrp> grpCleanup(grpData);
-	RsItem* item = nullptr;
-	RsTlvBinaryData& data = grpData->grp;
-
-	if (data.bin_len != 0)
-	{
-		// Create a local serialiser instance for deserializing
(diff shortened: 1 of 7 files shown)


src/services/p3wiki.cc
```diff
@@ -229,6 +293,294 @@ bool p3Wiki::getCollections(const std::list<RsGxsGroupId> groupIds, std::vector<
...
+bool p3Wiki::getModerators(const RsGxsGroupId& grpId, std::list<RsGxsId>& moderators)
+{
+	std::vector<RsWikiCollection> collections;
+	if (!getCollections({grpId}, collections) || collections.empty())
+		return false;
+
+	moderators.clear();
+	for (const auto& entry : collections.front().mModeratorTerminationDates)
+	{
+		if (entry.second == 0)
```
csoler: i really don't get what 0 means. Aren't you supposed to return all entries for which time(null) < entry.second ?

csoler (commented):
Good code. Make sure to check all the comments.

defnax:
Hi think the changes in rsgenechange.h and git ignore not needed

PR #256, opened 2026-01-30
Fix TorControlSocket shutdown hang by adding timeout

MHT417:
TorControlSocket::process() has an infinite loop that doesn't check the 
thread stop flag, this causes the thread to loop forever 

Added timeout check to allow clean thread termination.

### DIFF
--- a/src/tor/TorControl.cpp
+++ b/src/tor/TorControl.cpp
@@ -535,16 +535,28 @@ void TorControl::shutdown()
 
 void TorControl::shutdownSync()
 {
+
     if (!hasOwnership()) {
         RsWarn() << "torctrl: Ignoring shutdown command for a tor instance I don't own";
         return;
     }
-
     shutdown();
-    while (mSocket->moretowrite(0))
+    
+    // Add timeout to prevent infinite wait
+    int timeout_cnt = 0;
+    const int MAX_WAIT_ITERATIONS = 50; // 5 seconds max (50 * 100ms)
+    
+    while (mSocket->moretowrite(0) && timeout_cnt < MAX_WAIT_ITERATIONS) {
         std::this_thread::sleep_for(std::chrono::milliseconds(100));
-
+        timeout_cnt++;
+    }
+    
+    if (timeout_count >= MAX_WAIT_ITERATIONS) {
+        RsWarn() << "torctrl: Timeout waiting for socket to flush during shutdown, forcing close";
+    }
+    
     mSocket->close();
+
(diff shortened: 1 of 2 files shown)


csoler (commented):
One question pending.

jolavillette:
We are hitting this in the field on a Tor hidden node running a recent build (libretroshare a9ffc11eb, 2026-09-06): closing the GUI hangs in `RsTor::stop()` with `waitWhileStopping() ... TorManager to stop` climbing past 100 s, while the rest of the node keeps running (GXS sync, incoming SSL connections), since `rsGlobalShutDown()` only runs after `RsTor::stop()` returns. The user ends up killing the process, which also loses everything the config manager saves at exit.

The mechanism matches this PR exactly:

1. `TorControlSocket::process()` handles a partial line (bytes in `in_buffer` without `\n`) by sleeping 50 ms and looping. But the only thing that can complete that line is `read_pending()`, which runs from `tick()` on the same thread, and `tick()` is blocked inside `process()`. So once a control reply or event is split across two `read()` calls (1025-byte chunks in `RsFdBinInterface::read_pending()`), the socket thread spins forever. Nothing is visible while RS runs, since tor itself keeps working.
2. At shutdown, `TorControl::shutdownSync()` queues `SIGNAL SHUTDOWN` and waits for `moretowrite()` to clear. The write happens in `tick()` on the stuck thread, so it never does, and `TorManager::run()` never returns.

About the question "what if the network is just slow": returning from `process()` loses nothing. The partial line stays in `in_buffer`, the next `tick()` reads the rest, and `process()` resumes from where it was. In the current code the wait cannot succeed by construction, because the thread that waits is the one that would have to read. So the timeout only breaks a self-deadlock; a plain `return` instead of the sleep-and-continue would even be enough, with the 5 s bound in `shutdownSync()` as a safety net.

Log excerpt from the user:

```
D 1789113942.123 void RsThread::waitWhileStopping() 71 seconds passed waiting for thread: 7f011ddf06c0 TorManager to stop
...
I 1789113951.431 int pqissl::accept_locked(SSL*, int, const sockaddr_storage&) SUCCESSFUL connection to: 74d5c68b58ad11b8cad7c035e0bffc4a remoteaddr: 127.0.0.1
...
D 1789113985.146 void RsThread::waitWhileStopping() 114 seconds passed waiting for thread: 7f011ddf06c0 TorManager to stop
```

It would be great to have this merged; we will ship it in our builds in the meantime.

PR #273, opened 2026-02-26
fix macOS crashes in TorManager (invalid SYS_gettid syscall) and bytearray (memcpy block on null pointers)

jolavillette:
### Description

This PR fixes two critical, distinct bugs that cause RetroShare to crash on macOS during the Tor hidden service generation and initialization processes. Both issues stem from macOS having stricter system-level constraints and memory handling than Linux.

#### 1. Fix SIGABRT in [TorManager.cpp] (Invalid sys-call)
- **The Issue:** `RsTor::instance()` used `syscall(SYS_gettid)` to verify it was running on the main thread. However, `SYS_gettid` is a Linux-only system call that does not exist on macOS (Darwin/XNU). This caused the `assert()` to fail (`getpid() == -1`), resulting in a mandatory `SIGABRT` crash.
- **The Fix:** Replaced the OS-dependent system call with a cross-platform check using `std::this_thread::get_id()`. The thread ID is now securely cached upon the first invocation and correctly asserts against subsequent calls on all platforms.

#### 2. Fix SIGSEGV / Undefined Behavior in [bytearray.h] (Null pointer exceptions)
- **The Issue:** [ByteArray] operations (`memcpy`, `memcmp`, `strncmp`) were being called on empty strings (`size() == 0`), passing `nullptr` to these low-level C functions. While Linux glibc silently ignores `memcpy(nullptr, nullptr, 0)`, macOS clang/libc++ aggressively flags this as Undefined Behavior and triggers a hard crash.
- **The Fix:** Added strict `size > 0` and `!empty()` guard clauses before all memory operations within [bytearray.h]. Additionally, patched an off-by-one truncation bug in `ByteArray::replace` that inadvertently stripped trailing characters.

### DIFF
--- a/src/tor/TorManager.cpp
+++ b/src/tor/TorManager.cpp
@@ -154,11 +154,15 @@ std::string TorManager::torDataDirectory() const
 
 void TorManager::setTorDataDirectory(const std::string &path)
 {
-    assert(RsDirUtil::checkCreateDirectory(std::string(path)));
+    if(!RsDirUtil::checkCreateDirectory(path))
+    {
+        RsErr() << "TorManager::setTorDataDirectory() cannot create directory: " << path ;
+        return ;
+    }
 
     d->dataDir = path;
 
-    if (!d->dataDir.empty() && !ByteArray(d->dataDir).endsWith('/'))
+    if (!d->dataDir.empty() && !(d->dataDir.back() == '/'))
         d->dataDir += '/';
 }
 
@@ -937,10 +941,18 @@ void RsTor::setHiddenServiceDirectory(const std::string& dir)
     instance()->setHiddenServiceDirectory(dir);
 }
 
+#ifdef __APPLE__
+#include <pthread.h>
+#endif
+
 TorManager *RsTor::instance()
 {
+#ifdef __APPLE__
+    assert(pthread_main_np() != 0); // On macOS, ensure we are on the main thread
+#else
     static std::thread::id main_thread_id = std::this_thread::get_id();
     assert(std::this_thread::get_id() == main_thread_id); // make sure we're not in a different thread
+#endif
 
     if(rsTor == nullptr)
         rsTor = new TorManager;
--- a/src/tor/TorManager.cpp
+++ b/src/tor/TorManager.cpp
@@ -34,12 +34,6 @@
 #include <fstream>
 #include <stdio.h>
 
-// This works on linux only. I have no clue how to do that on windows. Anyway, this
-// is only needed for an assert that should normaly never be triggered.
-
-#if !defined(_WIN32) && !defined(__MINGW32__)
-#include <sys/syscall.h>
-#endif
 
 #include "util/rsdir.h"
 #include "retroshare/rsinit.h"
--- a/src/tor/TorManager.cpp
+++ b/src/tor/TorManager.cpp
@@ -43,6 +43,7 @@
 
 #include "util/rsdir.h"
 #include "retroshare/rsinit.h"
+#include <thread>
 
 #include "TorManager.h"
 #include "TorProcess.h"
@@ -938,9 +939,8 @@ void RsTor::setHiddenServiceDirectory(const std::string& dir)
 
 TorManager *RsTor::instance()
 {
-#if !defined(_WIN32) && !defined(__MINGW32__)
-    assert(getpid() == syscall(SYS_gettid));// make sure we're not in a thread
-#endif
+    static std::thread::id main_thread_id = std::this_thread::get_id();
+    assert(std::this_thread::get_id() == main_thread_id); // make sure we're not in a different thread
 
     if(rsTor == nullptr)
         rsTor = new TorManager;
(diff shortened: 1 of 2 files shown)


src/tor/bytearray.h
```diff
@@ -18,11 +18,11 @@ class ByteArray: public std::vector<unsigned char>
 public:
     ByteArray() =default;
     explicit ByteArray(int n) : std::vector<unsigned char>(n) {}
-    explicit ByteArray(const unsigned char *d,int n) : std::vector<unsigned char>(n) { memcpy(data(),d,n); }
+    explicit ByteArray(const unsigned char *d,int n) : std::vector<unsigned char>(n) { if (n > 0 && d) memcpy(data(),d,n); }
```
csoler: Because it is a constructor, I would actually prefer that you put an assert(n>0) here, because it would help properly catching earlier some situations that shouldn't happen anyway. The way it's done in the PR, the ByteArray silently ignores them.

src/tor/bytearray.h
```diff
@@ -18,11 +18,11 @@ class ByteArray: public std::vector<unsigned char>
...
     virtual ~ByteArray() =default;
 
-    ByteArray(const std::string& c) { resize(c.size()); memcpy(data(),c.c_str(),c.size()); }
-    const ByteArray& operator=(const std::string& c) { resize(c.size()); memcpy(data(),c.c_str(),c.size()); return *this; }
+    ByteArray(const std::string& c) { resize(c.size()); if(c.size() > 0) memcpy(data(),c.c_str(),c.size()); }
```
csoler: same here.

src/tor/bytearray.h
```diff
@@ -69,16 +69,19 @@ class ByteArray: public std::vector<unsigned char>
...
     }
 
-    ByteArray mid(uint32_t n,int s=-1) const
+    ByteArray mid(uint32_t n, int s = -1) const
     {
-        ByteArray res((s>=0)?s:(size()-n));
-        memcpy(res.data(),&data()[n],res.size());
+        if (n >= size()) return ByteArray();
+        uint32_t max_len = (uint32_t)size() - n;
+        uint32_t len = (s >= 0) ? std::min((uint32_t)s, max_len) : max_len;
```
csoler, line 82: good catch!

src/tor/bytearray.h
```diff
@@ -108,6 +112,9 @@ class ByteArray: public std::vector<unsigned char>
             else
                 res.push_back((*this)[i++]);
 
+        for(;i<size();++i)
```
csoler, line 115: good catch too ;-)

src/tor/TorManager.cpp
```diff
@@ -159,11 +154,15 @@ std::string TorManager::torDataDirectory() const
 
 void TorManager::setTorDataDirectory(const std::string &path)
 {
-    assert(RsDirUtil::checkCreateDirectory(std::string(path)));
+    if(!RsDirUtil::checkCreateDirectory(path))
```
csoler, line 163: If the tor data dir cannot be accessed/created then the software should stop. You may display a message if needed, but there's really no point in continuing if that step fails.

src/tor/TorManager.cpp
```diff
@@ -936,10 +941,17 @@ void RsTor::setHiddenServiceDirectory(const std::string& dir)
...
+#endif
+
 TorManager *RsTor::instance()
 {
-#if !defined(_WIN32) && !defined(__MINGW32__)
-    assert(getpid() == syscall(SYS_gettid));// make sure we're not in a thread
+#ifdef __APPLE__
+    assert(pthread_main_np() != 0); // On macOS, ensure we are on the main thread
+#else
+    static std::thread::id main_thread_id = std::this_thread::get_id();
```
csoler: no! A static in a function will only be initialized the first time you call that function, so here main_thread_id is not necessarily the main thread.

csoler (changes requested):
Some nice errors removed, and a few things to change.

src/tor/bytearray.h
jolavillette: Added assert(n >= 0 && (n == 0 || d != nullptr)). This enforces the API contract in debug mode: if a size is specified, the pointer must not be null, helping catch logic errors during development.

src/tor/bytearray.h
jolavillette: Added an assertion to ensure that for any non-empty string, the internal pointer is valid before the copy, preventing potential undefined behavior.

src/tor/TorManager.cpp
jolavillette: Fixed. The software now uses RsFatal() and terminates the process (exit(1)) if the Tor data directory cannot be created. This follows the "fail-fast" principle to avoid running in a silently broken state.

src/tor/TorManager.cpp
jolavillette: Fixed. Replaced the flawed static thread ID initialization with robust platform-specific calls: pthread_main_np() on macOS and getpid() == gettid() (via syscall) on Linux, ensuring the TorManager instance is only accessed from the main thread.

jolavillette:
All comments taken into account. Compiles OK on linux. I can't test the behaviour on macOS for the moment.

PR #275, opened 2026-02-28
fix CMake build with RNP

dbear496:
Fixes CMake builds when using librnp.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,25 +315,27 @@ if(RS_RNPLIB)
 	)
 else()
 
-set(OPENPGPSDK_DEVEL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../openpgpsdk/")
-if(EXISTS "${OPENPGPSDK_DEVEL_DIR}/CMakeLists.txt" )
-	message(
-		STATUS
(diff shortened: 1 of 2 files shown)


csoler:
I would be better to have RNP as the default and openpgp-sdk as an option, since RS has officially switched to rnp now.

G10h4ck:
@csoler @dbear496 to make it default just add a block like this ( in alphabetical order) in topdir `CMakeLists.txt`

```
option(
	RS_RNPLIB
	"Use RNP for PGP operations"
	ON )
```

dbear496:
Okay, I defined the `RS_RNPLIB` option in libretroshare with the default `ON`. However, just a heads up that the default for `RS_RNPLIB` in retroshare-service is still `OFF`, so building retroshare-service will still default to not use rnp.

G10h4ck:
@dbear496 please fix retroshare-service too

dbear496:
retroshare-service is not part of this repository.

G10h4ck:
@dbear496 yes, please open a PR on that repository too. @csoler anything else to change before merging?

csoler:
Does RS_RNP_LIB also works as a preprocessor variable (question from a non cmake user) ? RS needs one to include the proper files.

dbear496:
@csoler , that's a good catch. I added the USE_OPENPGPSDK define in CMake.

PR #281, opened 2026-03-18
fix(file_sharing): prevent infinite loop on circular symlinks when duplicates are allowed

jolavillette:
fix(file_sharing): prevent infinite loop on circular symlinks when duplicates are allowed

### DIFF
--- a/src/file_sharing/directory_updater.cc
+++ b/src/file_sharing/directory_updater.cc
@@ -24,6 +24,7 @@
 
 #include "util/cxx17retrocompat.h"
 #include "util/folderiterator.h"
+#include <algorithm>
 #include "util/rstime.h"
 #include "rsserver/p3face.h"
 #include "directory_storage.h"
@@ -219,10 +220,13 @@ bool LocalDirectoryUpdater::sweepSharedDirectories(bool& some_files_not_ready)
 	{
 		RS_DBG4("recursing into \"", stored_dir_it.name());
 
-		existing_dirs.insert(RsDirUtil::removeSymLinks(stored_dir_it.name()));
+		std::string canonical = RsDirUtil::removeSymLinks(stored_dir_it.name());
+		existing_dirs.insert(canonical);
+		std::vector<std::string> current_branch_real_paths;
+		current_branch_real_paths.push_back(canonical);
 		recursUpdateSharedDir(
 		            stored_dir_it.name(), *stored_dir_it,
-		            existing_dirs, 1, some_files_not_ready );
+		            existing_dirs, current_branch_real_paths, 1, some_files_not_ready );
 		/* here we need to use the list that was stored, instead of the shared
 		 * dir list, because the two are not necessarily in the same order. */
 	}
@@ -240,7 +244,7 @@ bool LocalDirectoryUpdater::sweepSharedDirectories(bool& some_files_not_ready)
 
 void LocalDirectoryUpdater::recursUpdateSharedDir(
         const std::string& cumulated_path, DirectoryStorage::EntryIndex indx,
-        std::set<std::string>& existing_directories, uint32_t current_depth,
+        std::set<std::string>& existing_directories, std::vector<std::string>& current_branch_real_paths, uint32_t current_depth,
         bool& some_files_not_ready )
 {
 	RS_DBG4("parsing directory \"", cumulated_path, "\" index: ", indx);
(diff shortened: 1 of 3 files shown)


src/file_sharing/directory_updater.cc
```diff
@@ -219,10 +220,13 @@ bool LocalDirectoryUpdater::sweepSharedDirectories(bool& some_files_not_ready)
 	{
 		RS_DBG4("recursing into \"", stored_dir_it.name());
 
-		existing_dirs.insert(RsDirUtil::removeSymLinks(stored_dir_it.name()));
+		std::string canonical = RsDirUtil::removeSymLinks(stored_dir_it.name());
```
csoler, line 222: when there is a loop in the sym links, canonicalize_file_name() will return nullptr and RsDirUtil::removeSymLinks() will return a null std::string (empty()=true). So inserting this empty string in existing_dirs has probably unwanted consequences.

src/file_sharing/directory_updater.cc
```diff
@@ -300,21 +304,31 @@ void LocalDirectoryUpdater::recursUpdateSharedDir(
...
 
-					if(dir_is_accepted && mFollowSymLinks && mIgnoreDuplicates)
+					if(dir_is_accepted && mFollowSymLinks)
 					{
 						std::string real_path = RsDirUtil::removeSymLinks(
 						            RsDirUtil::makePath(cumulated_path, dirIt.file_name()) );
 
-						if( existing_directories.end() !=
-						        existing_directories.find(real_path) )
+						if (std::find(current_branch_real_paths.begin(), current_branch_real_paths.end(), real_path) != current_branch_real_paths.end())
```
csoler: Bad idea to do a find on a std::vector. std::set would be wiser.

csoler (commented):
It seems to me that the AI has tried to re-do what we already do with existing_paths in a new structure (a std::vector, which is bad for find() btw), whereas the problem of circular paths is already handled. What may happen is that if there is a circular symlink, RsDirUtil::removeSymLinks will return an empty string, which is not accounter for. That's probably where the infinite loop comes from.

jolavillette:
AI says

Thanks for the review! Here's what this v2 addresses:

**1. [removeSymLinks] NULL crash — fixed ✅**

You were right: `canonicalize_file_name()` returns `NULL` on circular symlinks (ELOOP), and `std::string(nullptr)` is UB. Now guarded with a `nullptr` check, returning an empty string + `RS_WARN`.

**2. `std::vector` → `std::set` — fixed ✅**

Agreed, `std::find` on a vector was a bad choice. Replaced with `std::set` and `set::find()`.

**3. Why `current_branch_real_paths` is still needed**

The `existing_directories` set is only checked when `mIgnoreDuplicates == true` (old code: `if(dir_is_accepted && mFollowSymLinks && mIgnoreDuplicates)`). When the user disables "ignore duplicates", **no** cycle detection happens at all — the loop runs until depth 64, which is exponential (2^N paths with N symlinks).

`current_branch_real_paths` tracks only the **ancestors on the current DFS branch**, not all visited directories globally. This lets us:
- **Always** detect cycles (regardless of `mIgnoreDuplicates`)
- **Still allow** the same directory to appear in different branches when the user wants duplicates

The logic is now cleanly separated into 3 levels:
1. `real_path.empty()` → broken/circular symlink (ELOOP from `canonicalize_file_name`)
2. `current_branch_real_paths.find()` → symlink points to an ancestor (cycle)
3. `existing_directories.find()` → duplicate across branches (only if `mIgnoreDuplicates`)

src/file_sharing/directory_updater.cc
```diff
@@ -219,10 +219,15 @@ bool LocalDirectoryUpdater::sweepSharedDirectories(bool& some_files_not_ready)
 	{
 		RS_DBG4("recursing into \"", stored_dir_it.name());
 
-		existing_dirs.insert(RsDirUtil::removeSymLinks(stored_dir_it.name()));
+		std::string canonical = RsDirUtil::removeSymLinks(stored_dir_it.name());
+		if(!canonical.empty())
+			existing_dirs.insert(canonical);
+		std::set<std::string> current_branch_real_paths;
```
csoler: If canonical.empty() is true, it means the path (including the link) is not valid. So there's no point is continuing here. You can also factor the two ifs, which gives:
if(!canonical.empty())
{
    existing_dirs.insert(canonical);

    std::set<std::string> current_branch_real_paths = {canonical};

    recursUpdateSharedDir(
		            stored_dir_it.name(), *stored_dir_it,
		            existing_dirs, current_branch_real_paths, 1, some_files_not_ready );
}

src/file_sharing/directory_updater.cc
```diff
@@ -364,9 +385,19 @@ void LocalDirectoryUpdater::recursUpdateSharedDir(
 	// go through the list of sub-dirs and recursively update
 	for( DirectoryStorage::DirIterator stored_dir_it(mSharedDirectories, indx);
 	     stored_dir_it; ++stored_dir_it )
-		recursUpdateSharedDir( RsDirUtil::makePath(cumulated_path, stored_dir_it.name()),
-		                       *stored_dir_it, existing_directories,
+	{
+		std::string next_path = RsDirUtil::makePath(cumulated_path, stored_dir_it.name());
+		std::string canonical = RsDirUtil::removeSymLinks(next_path);
+		if(!canonical.empty())
+			current_branch_real_paths.insert(canonical);
```
csoler: Same here: if the path points to an ancestor, we should ignore it and return.

src/file_sharing/directory_updater.cc
```diff
@@ -364,9 +385,19 @@ void LocalDirectoryUpdater::recursUpdateSharedDir(
...
+
+		recursUpdateSharedDir( next_path,
+		                       *stored_dir_it, existing_directories, current_branch_real_paths,
 		                       current_depth+1, some_files_not_ready );
+
+		if(!canonical.empty())
```
csoler: ...so I suggest you replace all this by 

      current_branch_real_paths.insert(canonical);
      recursUpdateSharedDirs(...);
      current_branch_real_path.erase(canonical);

and add a comment that canonical cannot be empty.

csoler (changes requested):
So in short, we cannot trust the AI.

PR #284, opened 2026-04-12
Add admin signature validation in RsGenExchange

zapek:
It is possible to fool Retroshare into accepting a group without verification by creating one that has no signature at all (no admin signature and no author signature).

Now it will check the admin signature for every new group, not just group updates.

### DIFF
--- a/src/gxs/rsgenexchange.cc
+++ b/src/gxs/rsgenexchange.cc
@@ -1114,10 +1114,42 @@ int RsGenExchange::validateGrp(RsNxsGrp* grp)
     }
 
     if(idValidate)
+	{
+		// Validate admin signature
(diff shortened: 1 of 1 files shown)


zapek:
Maybe it's possible to write a common function with updateValid(), although they differ slightly. I did test my fix by modifying Xeres to send unsigned groups.

csoler:
well done! I think that if the admin key is not checked you can even receive a group and send a modified version of it to friends who do not have it yet => they will accept it.

csoler:
> Maybe it's possible to write a common function with updateValid(), although they differ slightly. I did test my fix by modifying Xeres to send unsigned groups.

In a second time yes. That kind of change needs to be super careful. Also generally a good strategy should be to never let "OK" be the default on signature checks.

PR #286, opened 2026-04-22
Fixed Use as direct source being automatically enabled when re-accepting a denied friend via Activity / Peer Details

jolavillette:
Fxed Use as direct source being automatically enabled when re-accepting a denied friend via Activity / Peer Details

Changed Default: Updated RS_NODE_PERM_DEFAULT to NONE in rspeers.h. The option is now unchecked by default when re-accepting a friend.
Adjusted Logic: Updated addFriend in p3peermgr.cc to skip permission masking for new or re-added friends. This ensures the new global default doesn't block explicit flag settings during initial setup.
Result: The checkbox no longer auto-enables via the Activity Feed, while remaining correctly configurable in the connection wizard.

### DIFF
--- a/src/pqi/p3peermgr.cc
+++ b/src/pqi/p3peermgr.cc
@@ -1069,7 +1069,14 @@ bool p3PeerMgrIMPL::addFriend(const RsPeerId& input_id, const RsPgpId& input_gpg
 		mLinkMgr->addFriend(id, vs_dht != RS_VS_DHT_OFF);
 	}
 
-	service_flags &= servicePermissionFlags(gpg_id) ; // Always reduce the permissions.
+	{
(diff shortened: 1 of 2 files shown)


csoler:
The addFriend function adds a friend with its SSL id (signed by pgp profile ID). The &= is here to make sure that you cannot change the flags for a profile (meaning all friends of that profile) by adding a new friend with more permissive flags. When the profile is not known, servicePermissionFlags(pgp_id) will return the default value (which is changed to NONE here). 

Only doing the &= when the profile  doesn't exist is not useful since it will default to NONE if it doesn't exist.

I believe that the behavior of the UI is simply changed in the first place after setting the default service flags to NONE.

PR #290, opened 2026-04-27
Added export & import mails

defnax:


csoler:
Messages are sent to GxsIds (they can also be send to nodes, but this needs to be removed), so because a GxsId is hosted to a single node, I don't see the point in exporting msgs you received for a given GxsId as new messages you would have received on a different node, since the destination ID is not there (and therefore cannot even respond).

PR #295, opened 2026-05-06
gxs: Fix multi-admin forum validation using fallback public keys

jolavillette:
gxs: Fix multi-admin forum validation using fallback public keys

GXS Forum Synchronization Failure on Multi-Admin Groups

Origin of the Issue:
The issue was introduced by a recent security hardening change implemented recently. To prevent unauthorized modifications to group metadata and protect against malicious key injections, the security reinforcement made administrator signature validation (INDEX_AUTHEN_ADMIN) extremely strict.

However, this hardening was written under the assumption that a GXS group only ever relies on a single primary administrator key matching the Group ID (metaData.mGroupId). This strict single-key assumption broke backward compatibility with legacy and multi-administrator forums (such as Retroshare Key Exchanges and RetroShare Plugin Ideas), which utilize multiple valid administrative keys stored in the public_keys map. Modern nodes running this hardened validation were thus rejecting these groups, causing them to fail synchronization on those nodes.

Problem Description:
When a multi-admin group arrived, RsGenExchange::validateGrp() failed because the administrator's cryptographic signature was validated strictly against the primary group ID key. If that specific main key failed or was missing, the validation failed completely (VALIDATE_FAIL), even though other valid administrator public keys were present in the metadata's public_keys map.

Resolution:
We implemented a secure cryptographic fallback inside RsGenExchange::validateGrp(). If the administrator's signature validation fails using the primary key, the engine now iterates through the alternative keys found in the public_keys map. If any fallback key successfully verifies the signature, the validation succeeds (VALIDATE_SUCCESS).

This preserves the recent security hardening while restoring full interoperability and synchronization for legacy and multi-admin forums.

### DIFF
--- a/src/gxs/rsgenexchange.cc
+++ b/src/gxs/rsgenexchange.cc
@@ -30,6 +30,7 @@
 #include "util/contentvalue.h"
 #include "util/rsprint.h"
 #include "util/rstime.h"
+#include "util/rsdebug.h"
 #include "retroshare/rsgxsflags.h"
 #include "retroshare/rsgxscircles.h"
 #include "retroshare/rsgrouter.h"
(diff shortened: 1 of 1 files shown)


zapek:
And another attempt at weakening the encryption. Are you working for the Mossad or something?

Again, each group is created with an admin key and signed by it. If the validation fails, the group is invalid. As simple as that.

Oh and guess where the Group ID comes from? The admin key ID. There's no such thing as a "multi-admin forums". The admins are a list of GXS IDs stored in the forum group item itself.

csoler:
I agree with Zapek. Wrong analysis from the IA.

PR #298, opened 2026-05-18
build(cmake): minimal fixes in CMakeLists.txt for transitive restbed headers, i2pcommon, and versioning

jolavillette:
build(cmake): minimal fixes in CMakeLists.txt for transitive restbed headers, i2pcommon, and versioning

This PR stabilizes the CMake configuration of libretroshare to support downstream modular builds (such as retroshare-gui and retroshare-service).

To ensure maximum stability and ease of review, the modifications made to the CMakeLists files are kept strictly minimal and highly targeted:

- Transitive Headers (1 line change): Promoted the restbed include directory to PUBLIC in CMakeLists.txt so that downstream modules cleanly inherit the HTTP/JSON API headers without duplicate include paths.
- Link Fixes (2 lines change): Reactivated the compilation of util/i2pcommon.cpp/.h under CMake, resolving undefined reference linking errors in downstream GUI settings panels.
- Version Display (minimal additions):
Directed the GUI human-readable version query to run in the parent directory's .git if present, ensuring the GUI gets the correct overall application version instead of submodule tags.
Added the RS_LIB_VERSION_HASH compile definition on the core target, fixing the [version not available] display bug in the about box.

All changes are non-intrusive and strictly limited to resolving compile/link-time blockers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -484,9 +484,9 @@ if(RS_JSON_API)
 		# dependencies see https://stackoverflow.com/a/64900982
 
 		target_include_directories(
-			${PROJECT_NAME} PRIVATE "${RESTBED_DEVEL_DIR}/source/" )
+			${PROJECT_NAME} PUBLIC "${RESTBED_DEVEL_DIR}/source/" )
 		target_link_directories(${PROJECT_NAME} PUBLIC "${CMAKE_CURRENT_BINARY_DIR}/../supportlibs/restbed")
-		target_link_libraries(${PROJECT_NAME} PRIVATE restbed)
+		target_link_libraries(${PROJECT_NAME} PRIVATE restbed-static)
 	else()
 		message( WARNING
 		         "FetchContent restbed, will be installed if you run make install")
@@ -503,7 +503,7 @@ if(RS_JSON_API)
 		FetchContent_MakeAvailable(restbed)
 
 		target_include_directories(
-			${PROJECT_NAME} PRIVATE ${restbed_SOURCE_DIR}/source/ )
+			${PROJECT_NAME} PUBLIC ${restbed_SOURCE_DIR}/source/ )
 		target_link_libraries(${PROJECT_NAME} PRIVATE restbed-static)
 	endif()
 
@@ -690,9 +690,14 @@ else()
 	set(V_GIT_DESCRIBE_OUTPUT "")
 	set(V_GIT_DESCRIBE_REGEXP "^v([0-9]+).([0-9]+).([0-9]+)(.*)\n$")
 
+	set(V_GIT_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
+	if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/../.git")
+		set(V_GIT_DIR "${CMAKE_CURRENT_SOURCE_DIR}/..")
+	endif()
+
 	execute_process(
 		COMMAND ${GIT_EXECUTABLE} describe --tags --long --match v*.*.*
-		WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+		WORKING_DIRECTORY "${V_GIT_DIR}"
 		OUTPUT_VARIABLE V_GIT_DESCRIBE_OUTPUT )
 
 	if(V_GIT_DESCRIBE_OUTPUT MATCHES "${V_GIT_DESCRIBE_REGEXP}")
@@ -720,6 +725,22 @@ if(V_VERSION_SET)
 		PUBLIC RS_MINI_VERSION=${RS_MINI_VERSION}
 		PUBLIC RS_EXTRA_VERSION="${RS_EXTRA_VERSION}" )
 endif(V_VERSION_SET)
+
+# Define the libretroshare specific version hash (submodule hash)
+set(V_LIB_GIT_HASH "")
+execute_process(
+	COMMAND ${GIT_EXECUTABLE} describe --tags --abbrev=7 --always --dirty
+	WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+	OUTPUT_VARIABLE V_LIB_GIT_HASH
+	OUTPUT_STRIP_TRAILING_WHITESPACE )
+
+if(V_LIB_GIT_HASH STREQUAL "")
+	set(V_LIB_GIT_HASH "[version not available]")
+endif()
+
+target_compile_definitions(
+	${PROJECT_NAME}
+	PUBLIC RS_LIB_VERSION_HASH="${V_LIB_GIT_HASH}" )
 
 ################################################################################
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -692,7 +692,7 @@ endif(RS_ANDROID)
 
 ################################################################################
 
-set( CMAKE_CXX_FLAGS "-Wno-deprecated-declarations" )
+target_compile_options(${PROJECT_NAME} PRIVATE -Wno-deprecated-declarations)
 target_compile_definitions(${PROJECT_NAME} PUBLIC RS_NO_WARN_DEPRECATED )
 
 set(V_VERSION_SET OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,6 +245,7 @@ else()
 	set_property(TARGET ${PROJECT_NAME} PROPERTY CXX_STANDARD 17)
 endif()
 
+
 ## As of today libretroshare doesn't hide implementation details properly so it
 ## is necessary to flag all implementation headers as public
 target_include_directories(
@@ -299,8 +300,8 @@ target_link_libraries(${PROJECT_NAME} PRIVATE OpenSSL::SSL OpenSSL::Crypto)
 if(RS_RNPLIB)
 	# Add RNP build directories to linker search path
 	target_link_directories(${PROJECT_NAME} PRIVATE
-		${CMAKE_CURRENT_BINARY_DIR}/../supportlibs/librnp/Build/src/lib
-		${CMAKE_CURRENT_BINARY_DIR}/../supportlibs/librnp/Build/src/libsexpp
+		${CMAKE_CURRENT_SOURCE_DIR}/../supportlibs/librnp/Build/src/lib
+		${CMAKE_CURRENT_SOURCE_DIR}/../supportlibs/librnp/Build/src/libsexpp
 	)
 
 	# Link libraries by name, including RNP libs and dependencies first
@@ -311,7 +312,7 @@ if(RS_RNPLIB)
 
 	target_include_directories(${PROJECT_NAME} PUBLIC
 		"${CMAKE_CURRENT_SOURCE_DIR}/../supportlibs/librnp/include"
-		"${CMAKE_CURRENT_BINARY_DIR}/../supportlibs/librnp/Build/src/lib"
+		"${CMAKE_CURRENT_SOURCE_DIR}/../supportlibs/librnp/Build/src/lib"
 	)
 else()
 
@@ -430,19 +431,21 @@ if(RS_JSON_API)
 		JSONAPI_GENERATOR_EXECUTABLE
 		"${JSONAPI_GENERATOR_SOURCE_DIR}/jsonapi-generator.py" )
 
-	file(
-		COPY "src/jsonapi/jsonapi-generator-doxygen.conf"
-		DESTINATION "${JSON_API_GENERATOR_WORK_DIR}" )
+	if(NOT EXISTS "${JSON_API_GENERATOR_DOXYFILE}")
+		file(
+			COPY "src/jsonapi/jsonapi-generator-doxygen.conf"
+			DESTINATION "${JSON_API_GENERATOR_WORK_DIR}" )
 
-	file(
-		APPEND
-		"${JSON_API_GENERATOR_DOXYFILE}"
-		"OUTPUT_DIRECTORY=${JSONAPI_GENERATOR_OUTPUT_DIR}\n"
-		"INPUT=${CMAKE_CURRENT_SOURCE_DIR}/src/\n"
-		"EXCLUDE="
-		"	${CMAKE_CURRENT_SOURCE_DIR}/src/tests"
-		"	${CMAKE_CURRENT_SOURCE_DIR}/src/unused"
-		"	${CMAKE_CURRENT_SOURCE_DIR}/src/unfinished\n" )
+		file(
+			APPEND
+			"${JSON_API_GENERATOR_DOXYFILE}"
+			"OUTPUT_DIRECTORY=${JSONAPI_GENERATOR_OUTPUT_DIR}\n"
+			"INPUT=${CMAKE_CURRENT_SOURCE_DIR}/src/\n"
+			"EXCLUDE="
+			"	${CMAKE_CURRENT_SOURCE_DIR}/src/tests"
+			"	${CMAKE_CURRENT_SOURCE_DIR}/src/unused"
+			"	${CMAKE_CURRENT_SOURCE_DIR}/src/unfinished\n" )
+	endif()
 
 	add_custom_command(
 		OUTPUT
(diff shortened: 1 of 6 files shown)


src/jsonapi/jsonapi-generator.py
```diff
@@ -190,15 +190,17 @@ def processFile(file):
...
+				if pName in paramsMap:
+					tmpParam = paramsMap[pName]
+					tmpD = tmpPN.attrib['direction'] if 'direction' in tmpPN.attrib else ''
+
+					if 'in' in tmpD:
+						tmpParam._in = True
+						hasInput = True
+					if 'out' in tmpD:
+						tmpParam._out = True
+						hasOutput = True
```
G10h4ck, line 203: This change seems unrelated to the PR description, what it is supposed to do?

src/pqi/sslfns.h
```diff
@@ -1,4 +1,4 @@
-﻿/*******************************************************************************
+/*******************************************************************************
```
G10h4ck, line 1: The line ending was wrong?

src/pqi/sslfns.h
```diff
@@ -28,6 +28,13 @@
 
 /******************** notify of new Cert **************************/
 
+#ifdef X509_NAME
+#undef X509_NAME
+#endif
+#ifdef X509_EXTENSIONS
+#undef X509_EXTENSIONS
+#endif
+
```
G10h4ck, line 37: This change seems unrelated to the PR description, what it is supposed to do?

src/retroshare/rstor.h
```diff
@@ -41,6 +41,10 @@ enum class RsTorManagerEventCode: uint8_t
     TOR_MANAGER_STOPPED       = 0x06,
 };
 
+#ifdef ERROR
+#undef ERROR
+#endif
+
```
G10h4ck, line 47: This change seems unrelated to the PR description, what it is supposed to do?

src/util/rsdebug.h
```diff
@@ -38,6 +38,10 @@
 #include "util/rsmacrosugar.hpp"
 
 
+#ifdef ERROR
+#undef ERROR
+#endif
+
```
G10h4ck, line 44: This change seems unrelated to the PR description, what it is supposed to do?

CMakeLists.txt
```diff
@@ -101,7 +101,7 @@ cmake_dependent_option(
 option(
 	RS_WARN_DEPRECATED
 	"Print warning about RetroShare deprecated components usage during build"
-	ON )
+	OFF )
```
G10h4ck, line 104: Why changing this? Also it is unrelated to the PR description. I do prefer it to be kept ON by default so the usage of deprecated thing is more visible

CMakeLists.txt
```diff
@@ -297,21 +298,27 @@ target_link_libraries(${PROJECT_NAME} PRIVATE OpenSSL::SSL OpenSSL::Crypto)
...
 if(RS_RNPLIB)
+	find_library(BOTAN_LIBRARY NAMES botan botan-3 botan-2 libbotan-3 libbotan-2 REQUIRED)
+
 	# Add RNP build directories to linker search path
-	target_link_directories(${PROJECT_NAME} PRIVATE
-		${CMAKE_CURRENT_BINARY_DIR}/../supportlibs/librnp/Build/src/lib
-		${CMAKE_CURRENT_BINARY_DIR}/../supportlibs/librnp/Build/src/libsexpp
+	target_link_directories(${PROJECT_NAME} PUBLIC
+		${CMAKE_CURRENT_SOURCE_DIR}/../supportlibs/librnp/Build/src/lib
+		${CMAKE_CURRENT_SOURCE_DIR}/../supportlibs/librnp/Build/src/libsexpp
```
G10h4ck, line 306: Is the `Build` directory name always the same?

CMakeLists.txt
```diff
@@ -297,21 +298,27 @@ target_link_libraries(${PROJECT_NAME} PRIVATE OpenSSL::SSL OpenSSL::Crypto)
...
 	# Link libraries by name, including RNP libs and dependencies first
-	target_link_libraries(${PROJECT_NAME} PRIVATE
+	target_link_libraries(${PROJECT_NAME} PUBLIC
 		rnp             # Link by name
 		sexpp
+		${BOTAN_LIBRARY}
+		json-c
+		bz2
+		z
 	)
```
G10h4ck, line 317: Here you are changing the linking of all those libraries from private to public, that means that ALL symbols will be visible when linking against libretroshare, that is unwanted also it raise an error like "too many symbols" on some platforms. Also I think libretroshare has no reason to expose all those symbols, if some downstream user need them, they will need to link to those libraries by themselves.

CMakeLists.txt
```diff
@@ -430,19 +437,21 @@ if(RS_JSON_API)
...
-	file(
-		APPEND
-		"${JSON_API_GENERATOR_DOXYFILE}"
-		"OUTPUT_DIRECTORY=${JSONAPI_GENERATOR_OUTPUT_DIR}\n"
-		"INPUT=${CMAKE_CURRENT_SOURCE_DIR}/src/\n"
-		"EXCLUDE="
-		"	${CMAKE_CURRENT_SOURCE_DIR}/src/tests"
-		"	${CMAKE_CURRENT_SOURCE_DIR}/src/unused"
-		"	${CMAKE_CURRENT_SOURCE_DIR}/src/unfinished\n" )
+	if(NOT EXISTS "${JSON_API_GENERATOR_DOXYFILE}")
```
G10h4ck, line 440: Why the conditional? In which case do you want it to not be overridden by the updated one?

CMakeLists.txt
```diff
@@ -465,6 +474,8 @@ if(RS_JSON_API)
 
 	set(BUILD_TESTS OFF CACHE BOOL "Do not build restbed tests")
 	set(BUILD_SSL OFF CACHE BOOL "Do not build restbed SSL support")
+	set(BUILD_SHARED_LIBRARY OFF)
+	set(BUILD_SHARED_LIBS OFF)
```
G10h4ck, line 478: Are the two of them needed?

CMakeLists.txt
```diff
@@ -498,13 +509,18 @@ if(RS_JSON_API)
 			GIT_SHALLOW TRUE
 			GIT_PROGRESS TRUE
 			TIMEOUT 10
-#			EXCLUDE_FROM_ALL # Available only in CMake >= 3.28
+			PATCH_COMMAND sed -i -e "s|/wd4251||g" -e "s|VERSION 3.1.0|VERSION 3.5.0|g" CMakeLists.txt
```
G10h4ck, line 512: What is this line supposed to do? At first glange it is quite brittle to hot patching CMake files line that, but above all why?

G10h4ck (changes requested):
Good things intermixed with wrong stuff, @jolavillette please fix them and answhere the inline questions ;-)

src/jsonapi/jsonapi-generator.py
jolavillette: Thank you for reviewing this PR. Note that since I pushed it I managed to compile also on Windows and MacOS. I had to make a few more changes in some CMakeLists.txt files. I just pushed the new cross platform PR on github: see https://github.com/RetroShare/libretroshare/pull/304

Please understand that even if I did my best along the development and cross platform testing of this PR to closely monitor what was going on, the replies below were AI generated, 

About json-pi generator:
This fixes a crash of the JSON API generator with some Doxygen versions (especially on Windows/macOS). The original code accesses tmpPN.text directly, which returns None when the XML node contains sub-elements instead of plain text (behavior varies across Doxygen versions). Using getText(tmpPN) (which does "".join(e.itertext())) is more robust. The if pName in paramsMap guard prevents a KeyError if Doxygen generates unexpected <parametername> entries.

src/pqi/sslfns.h
jolavillette: Yes, there was trailing whitespace at end of line. Automatic cleanup by the editor. Cosmetic only, no functional impact.

src/pqi/sslfns.h
jolavillette: This is a required fix for Windows compilation. Windows headers (<wincrypt.h>, pulled in transitively via system headers) define X509_NAME and X509_EXTENSIONS as preprocessor macros, which conflicts with their use as type identifiers in function signatures declared in sslfns.h (e.g. SignX509Certificate(X509_NAME *issuer, ...)). Without these #undef, compilation fails on Windows/MSVC.

This is a classic Windows compatibility pattern, also used by OpenSSL itself.

src/retroshare/rstor.h
jolavillette: Same issue as previous point: on Windows, <wingdi.h> (included transitively via <windows.h>) does #define ERROR 0. This breaks enum class members using ERROR as an enumerator, such as RsTorHiddenServiceStatus::ERROR and RsTorConnectivityStatus::ERROR. The preprocessor replaces ERROR with 0, causing a compilation error.

src/util/rsdebug.h
jolavillette: Same as previous points. RsLoggerCategories::ERROR is broken by the Windows macro #define ERROR 0. The #undef is required for rsdebug.h to compile on Windows.

CMakeLists.txt
jolavillette: You're right, this change is not needed for the PR. It was a local development convenience to reduce warning noise during cross-platform build debugging.

CMakeLists.txt
jolavillette: The Build directory is created by the librnp build script (supportlibs/librnp) and its name is hardcoded in that script. It's the same name on all 3 platforms.

The change from CMAKE_CURRENT_BINARY_DIR to CMAKE_CURRENT_SOURCE_DIR is because librnp is built outside of libretroshare's CMake build tree (via a separate script), so its build artifacts live in the source tree, not in CMake's build directory.

CMakeLists.txt
jolavillette: Without this condition, every CMake invocation (even a simple cmake --build that triggers a reconfigure) re-executes file(APPEND ...) and duplicates the OUTPUT_DIRECTORY, INPUT, EXCLUDE lines in the Doxyfile. After N reconfigurations, the file contains N copies of those directives, which generates Doxygen warnings and can cause unexpected behavior.

The original file(COPY ...) + file(APPEND ...) runs at every configure step, not just the first time. The condition prevents accumulation.

CMakeLists.txt
jolavillette: BUILD_SHARED_LIBS is the standard CMake variable (honored by add_library() when no type is specified). BUILD_SHARED_LIBRARY is a custom variable used specifically by restbed's CMakeLists. Both are needed because restbed uses its own convention on top of the standard CMake one.

CMakeLists.txt
jolavillette: This line does two things:

Removes /wd4251: this is an MSVC flag (/wd = disable warning) that restbed puts in its CMakeLists. This flag causes an error with GCC/Clang (which don't recognize it). Since we use FetchContent on all platforms, it must be removed on non-MSVC platforms.
Changes VERSION 3.1.0 → VERSION 3.5.0: restbed's CMakeLists specifies cmake_minimum_required(VERSION 3.1.0), which triggers a deprecation warning with CMake ≥ 3.27 (old policies are deprecated). Bumping to 3.5.0 silences this warning.
You're right that it's brittle — it breaks if restbed changes those exact strings in an update.

CMakeLists.txt
jolavillette: You are right, PUBLIC is too broad and exposes headers/symbols unnecessarily. The reason we switched to PUBLIC was to fix downstream link errors on Windows. However, the root cause wasn't PRIVATE itself (CMake does propagate PRIVATE static dependencies for linking), but rather the use of target_link_directories(PRIVATE), which didn't propagate the -L search paths to the downstream executables.

I will revert to PRIVATE. To fix the Windows link errors properly, I will remove target_link_directories and instead link directly to the absolute paths of these static libraries (e.g., ${CMAKE_CURRENT_SOURCE_DIR}/.../librnp.a). This allows CMake to correctly resolve the libraries for downstream consumers without polluting the public interface.

src/pqi/sslfns.h
G10h4ck: in that case the `#if` should be platform specific, like `#if def(WINDOWS) && def(X509_EXTENSIONS)` it also makes more explicit the intent of code, imagine reading that 2 years from now, surely you would exclaim WTF is this? also having some comment near this kind of stuff, makes it even more future readable. So please apply this adjustments to all the similar `#ifdef` introduced

src/retroshare/rstor.h
G10h4ck: see previous comment

src/util/rsdebug.h
G10h4ck: see previous comment

CMakeLists.txt
G10h4ck: ok, having a short comment about that would be helpful ;-)

CMakeLists.txt
G10h4ck: seems legit

CMakeLists.txt
G10h4ck: if the original is copied and then the directives appended every-time, how is it possible for the directives to accumulate?

CMakeLists.txt
G10h4ck: At least a comment is needed explaining what is going on, also an upstream PR is the right thing to do

G10h4ck:
> Thank you for reviewing this PR. Note that since I pushed it I managed to compile also on Windows and MacOS. I had to make a few more changes in some CMakeLists.txt files. I just pushed the new cross platform PR on github: see https://github.com/RetroShare/libretroshare/pull/304

@jolavillette should I keep reviewing this code or the #304 ? Please one thing at time so we avoid wasting code review time

CMakeLists.txt
G10h4ck: Is it necessary to hardcode disabling them? What if a downstream libretroshare user what to link dinamically? Having the linker linking also the .so should not harm.

jolavillette:
The changes following your comments are in 304, not 298.
I am sorry for the mess.
304 was built on top of 298.
It seemed reasonnable to me to proceed step by step and I chose to implemente the cmake migration for windows first and then macos.
I now realized that handling macos first would probably have been a better idea.
304 does not add many changes in libretroshare.
It would be better if you can switch to 304 from now.

src/pqi/sslfns.h
jolavillette: Done. I have wrapped the #undef inside #if defined(WINDOWS_SYS) && defined(X509_EXTENSIONS) to make it strictly platform-specific. I also added an explicit comment explaining that Windows headers (<wincrypt.h>) define this as a macro which conflicts with RetroShare types.

src/retroshare/rstor.h
jolavillette: Done. Similarly, I wrapped this inside #if defined(WINDOWS_SYS) && defined(ERROR) and added a comment explaining the conflict with <wingdi.h>.

src/util/rsdebug.h
jolavillette: Done. Same as the previous point, the #undef is now platform-specific with an explanatory comment.

CMakeLists.txt
jolavillette: Done. I have added a short comment explaining that the Build directory name is hardcoded in librnp's custom external build script.

CMakeLists.txt
jolavillette: Done. I reverted it to PRIVATE. To fix the downstream Windows link errors cleanly, I removed target_link_directories and instead linked directly to the absolute paths of the static libraries (e.g., librnp.a). This prevents exposing symbols unnecessarily.

CMakeLists.txt
jolavillette: The accumulation happens because CMake's file(COPY) skips the copy if the destination file is newer. Since file(APPEND) modifies the destination, its timestamp becomes newer than the original. On the next CMake run, COPY is silently skipped, but APPEND runs again, duplicating the text. To fix this robustly, I have removed the if(NOT EXISTS) check and completely refactored the logic to use file(READ), followed by a string(APPEND) in memory, and finally a single file(WRITE). This prevents any accumulation.

CMakeLists.txt
jolavillette: Done. I have removed the hardcoded OFF values. They are now dynamically set to mirror RetroShare's library preference (${RS_LIBRETROSHARE_SHARED}). Downstream users can now link dynamically if they want.

CMakeLists.txt
jolavillette: Done. I have added an explanatory comment above the command clarifying that the sed patch removes the MSVC-specific /wd4251 flag (which breaks GCC/Clang builds) and bumps the CMake VERSION to 3.5.0 to suppress deprecation warnings. I will also submit an upstream PR to restbed separately to fix this at the source.

G10h4ck:
should this be closed then ?

PR #301, opened 2026-05-24
fix(mail): avoid unconditional periodic saves of email database

jolavillette:
fix(mail): avoid unconditional periodic saves of email database

Problem Description
The periodic method p3MsgService::checkOutgoingMessages() is executed automatically every 5 seconds (via the service's tick() routine). A performance bug was originally introduced by libretroshare commit f2724f97 ("fix: save config after sending node-to-node msgs to avoid resending"). That commit aimed to prevent node-to-node messages from being sent repeatedly by forcing a configuration save immediately after they were transmitted. However, by adding the call to IndicateConfigChanged(SAVE_NOW) unconditionally at the end of checkOutgoingMessages(), it caused the function to trigger a disk write every 5 seconds, even when the outgoing queue was completely idle.

Applied Fix
This PR resolves this by making the call to IndicateConfigChanged(SAVE_NOW) conditional on a changed boolean flag. Disk writes are now only triggered when an actual state change occurs within the queue (e.g., when a message is successfully sent, fails, or is updated), preserving the original fix without the I/O penalty.

Audit of the 24 Other Occurrences of IndicateConfigChanged
Beside the fix a thorough audit of the other 24 calls to IndicateConfigChanged inside p3msgservice.cc was conducted.
All 24 calls are relevant and correct. Unlike the one in the periodic loop, they are strictly bound to explicit, database-altering actions (e.g., deleting a message, marking emails as read/unread, or changing tags).
Two of these calls are executed closely together during message deletion. However, they are not critical because IndicateConfigChanged only sets an in-memory dirty flag. The actual disk serialization is handled asynchronously by the configuration manager (p3ConfigMgr), avoiding redundant writes. As a result, it was deemed unnecessary to modify them.

### DIFF
--- a/src/services/p3msgservice.cc
+++ b/src/services/p3msgservice.cc
@@ -526,7 +526,7 @@ int p3MsgService::checkOutgoingMessages()
 
     if(changed)
     {
-        RsDbg() << "MAIL: checkOutgoingMessages() triggering IndicateConfigChanged(SAVE_NOW)" << std::endl;
+        // RsDbg() << "MAIL: checkOutgoingMessages() triggering IndicateConfigChanged(SAVE_NOW)" << std::endl;
         IndicateConfigChanged(RsConfigMgr::CheckPriority::SAVE_NOW);
     }
 
--- a/src/services/p3msgservice.cc
+++ b/src/services/p3msgservice.cc
@@ -395,6 +395,7 @@ void p3MsgService::checkSizeAndSendMessage(RsMsgItem *msg,const RsPeerId& destin
 
 int p3MsgService::checkOutgoingMessages()
 {
+    bool changed = false;
     auto pEvent = std::make_shared<RsMailStatusEvent>();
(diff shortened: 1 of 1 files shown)


csoler:
Indeed, the frequent saving of cfg files is not the source of the corruption of config files, but it makes it happen more often. The root cause is in the disk space safety limit not being correlated with the actual size of the saved files. The disk space limit is 100MB by default and .cfg files of such a size are very unlikely. To prevent them being corrupted, the saving in p3cfgmgr should be made more robust:
1 - pqistore.cc:418, return sizeItems==enc_bio->senddata(data,sizeItems);
2 - p3cfgmgr.cc:342: add a return in the if (after cleaning up, etc). For now, if the disk is full, it continues, and the signature will fail => the file gets corrupted and discarded at the next start.
3 - p3cfgmgr.cc: 361, add an if over writetofile. On error, quit with an error message.

PR #304, opened 2026-06-02
CMake migration modular cross platform

jolavillette:
Implement full modular cmake compilation on linux, windows and macos

Works with the following PRs in RetroShare and i2p/libsam3
https://github.com/RetroShare/RetroShare/pull/3229
https://github.com/i2p/libsam3/pull/30

Example of script to compile everything:

#!/bin/bash

# Exit immediately if any command fails
set -e

# Clean build directories if the "clean" argument is passed
if [ "$1" == "clean" ]; then
    echo "=== Cleaning build directories ==="
    rm -rf libretroshare/Build
    rm -rf retroshare-service/Build
    rm -rf retroshare-friendserver/Build
    rm -rf retroshare-gui/Build
fi

# Detect number of CPU cores for fast parallel compilation
NPROC=$(sysctl -n hw.ncpu 2>/dev/null || nproc 2>/dev/null || echo 4)

# Detect Qt5 path on macOS if Homebrew is installed
QT_PREFIX=""
if command -v brew &> /dev/null; then
    QT_PREFIX=$(brew --prefix qt@5)
fi

echo "=== Starting RetroShare Compilation ==="

# 1. Compile libretroshare (core library) and librnp automatically
echo ">>> Step 1: Compiling libretroshare..."
mkdir -p libretroshare/Build && cd libretroshare/Build
cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DRS_LIBRETROSHARE_STATIC=OFF -DRS_LIBRETROSHARE_SHARED=ON ..
cmake --build . -j${NPROC}
cd ../..

# 2. Compile retroshare-service (headless daemon)
echo ">>> Step 2: Compiling retroshare-service..."
mkdir -p retroshare-service/Build && cd retroshare-service/Build
cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 ..
cmake --build . -j${NPROC}
cd ../..

# 3. Compile retroshare-friendserver
echo ">>> Step 3: Compiling retroshare-friendserver..."
mkdir -p retroshare-friendserver/Build && cd retroshare-friendserver/Build
cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 ..
cmake --build . -j${NPROC}
cd ../..

# 4. Compile retroshare-gui (graphical user interface)
echo ">>> Step 4: Compiling retroshare-gui..."
mkdir -p retroshare-gui/Build && cd retroshare-gui/Build
cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
      -DCMAKE_PREFIX_PATH="${QT_PREFIX}" \
      -DENABLE_GUI=ON \
      -DENABLE_QT_TRANSLATIONS=ON ..
cmake --build . -j${NPROC}
cd ../..

echo "=== Compilation Successfully Completed! ==="

The PR also provides an orchestrator
See  "make help"


G10h4ck:
libnrp does use CMake too? In that case it shouldn't be needed to compile it manually before, if added properly as a dependency CMake will deal with compiling of that too automatically just like the other dependencies (ex: restbed)

src/pqi/sslfns.h
```diff
@@ -28,6 +28,15 @@
 
 /******************** notify of new Cert **************************/
 
+#if defined(WINDOWS_SYS) && defined(X509_NAME)
+// Workaround: Windows headers (<wincrypt.h>) define X509_NAME which conflicts with RetroShare types.
```
G10h4ck: Is RetroShare which define `X509_NAME` as a type? In that case RS shouldn't pollute the namespace like that, and the type should be renamed to `RS_X509_NAME`

src/pqi/sslfns.h
```diff
@@ -28,6 +28,15 @@
...
+#undef X509_NAME
+#endif
+#if defined(WINDOWS_SYS) && defined(X509_EXTENSIONS)
+// Workaround: Windows headers (<wincrypt.h>) define X509_EXTENSIONS which conflicts with RetroShare types.
```
G10h4ck: See previous comment

src/retroshare/rstor.h
```diff
@@ -41,6 +41,11 @@ enum class RsTorManagerEventCode: uint8_t
     TOR_MANAGER_STOPPED       = 0x06,
 };
 
+#if defined(WINDOWS_SYS) && defined(ERROR)
+// Workaround: Windows headers (<wingdi.h>) define ERROR as 0, which breaks enum members named ERROR.
```
G10h4ck: Being this a public header I think we should not mess with undefining a windows symbol, because that would happen also for any library downstream user. It is strange this is happening bacause the enum is scoped `enum class`. If failing anyway better investigating where the actual error is happening `rstor.h` doesn't include `<wingdi.h>` directly so it should not happen there. Another workaround could be to rename the enum values, but that should be a last resort.

src/util/rsdebug.h
```diff
@@ -38,6 +38,11 @@
 #include "util/rsmacrosugar.hpp"
 
 
+#if defined(WINDOWS_SYS) && defined(ERROR)
+// Workaround: Windows headers (<wingdi.h>) define ERROR as 0, which breaks enum members named ERROR.
+#undef ERROR
```
G10h4ck: Being this someway internal implementation here is less dangerous, but still why this used to work and now it doesn't? I remember to compile on windows a general thing must be set, it was called something "Windows lean and mean headers" or something like that so it stop polluting all around, please check that instead of undeffing things around, which is quite ugly and be done just as a last resort.

CMakeLists.txt
```diff
@@ -227,6 +227,13 @@ if(RS_LIBRETROSHARE_SHARED)
 	## Ensure statically linked libraries such as openpgpsdk are compiled with
 	## PIC Which is needed for shared library
 	set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
+	if(WIN32)
+		set_target_properties(${PROJECT_NAME} PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)
+		if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+			target_link_options(${PROJECT_NAME} PRIVATE "-Wl,--export-all-symbols,--exclude-libs,libcrypto.a:libssl.a:libsqlite3.a:libopenpgpsdk.a:libbitdht.a:librestbed.a:librestbed-static.a:libminiupnpc.a:librnp.a:libsexpp.a:libbotan.a:libbotan-3.a:libbotan-2.a:libz.a:libbz2.a:libjson-c.a")
```
G10h4ck: This doesn't smell good at all, hardcoding all those library here, is basically unmaintenable code, CMake is here to do this kind of things automatically, so if a new library get added it automatically sets the proper flags, same goes if a library dependency get removed

CMakeLists.txt
```diff
@@ -245,6 +252,19 @@ else()
 	set_property(TARGET ${PROJECT_NAME} PROPERTY CXX_STANDARD 17)
 endif()
 
+if(WIN32)
+	target_compile_definitions(
+		${PROJECT_NAME} PUBLIC
```
G10h4ck, line 267: Are you sure `PUBLIC` is needed here? Downstream user shouldn't be influenced by internal libretroshare flag choiche.

CMakeLists.txt
```diff
@@ -245,6 +252,19 @@ else()
...
+		WINDOWS_SYS
+		UNICODE
+		_UNICODE
+		NOGDI
+		NOMINMAX
+		NOCRYPT
```
G10h4ck, line 277: Here is where you need to add the "lean and mean" thing AFAIR, also check one by one those flags and check they are really needed, and what each of them does (add comments if not obvious)

CMakeLists.txt
```diff
@@ -297,21 +317,42 @@ target_link_libraries(${PROJECT_NAME} PRIVATE OpenSSL::SSL OpenSSL::Crypto)
 ################################################################################
 
 if(RS_RNPLIB)
-	# Add RNP build directories to linker search path
-	target_link_directories(${PROJECT_NAME} PRIVATE
-		${CMAKE_CURRENT_BINARY_DIR}/../supportlibs/librnp/Build/src/lib
-		${CMAKE_CURRENT_BINARY_DIR}/../supportlibs/librnp/Build/src/libsexpp
-	)
+	find_library(BOTAN_LIBRARY NAMES botan botan-3 botan-2 libbotan-3 libbotan-2 REQUIRED)
```
G10h4ck: The whole RNPLIB section smell quite bad, look at how I did add OPENPGPSDK_DEVEL_DIR and RAPIDJSON_DEVEL_DIR for examples on how to do this properly

CMakeLists.txt
```diff
@@ -473,7 +516,10 @@ if(RS_JSON_API)
 		## should be sent upstream.
 		## Restbed is therefore compiled by Android toolchain preparation script
 		target_link_libraries(${PROJECT_NAME} PRIVATE restbed)
-	elseif(EXISTS "${RESTBED_DEVEL_DIR}/CMakeLists.txt")
+	elseif(NOT WIN32 AND EXISTS "${RESTBED_DEVEL_DIR}/CMakeLists.txt")
+		## On Windows, skip the local submodule: recent restbed HEAD requires
+		## system ASIO (not available by default on MSYS2/MinGW). Fall through
+		## to FetchContent which downloads restbed 4.8 with bundled ASIO.
```
G10h4ck, line 520: This is a good way to making the code future readable ;-)

CMakeLists.txt
```diff
@@ -484,9 +530,9 @@ if(RS_JSON_API)
 		# dependencies see https://stackoverflow.com/a/64900982
 
 		target_include_directories(
-			${PROJECT_NAME} PRIVATE "${RESTBED_DEVEL_DIR}/source/" )
+			${PROJECT_NAME} PUBLIC "${RESTBED_DEVEL_DIR}/source/" )
 		target_link_directories(${PROJECT_NAME} PUBLIC "${CMAKE_CURRENT_BINARY_DIR}/../supportlibs/restbed")
-		target_link_libraries(${PROJECT_NAME} PRIVATE restbed)
+		target_link_libraries(${PROJECT_NAME} PRIVATE restbed-static)
```
G10h4ck: This sounds it will breaks all the targets that where working before (tens of Linux distributions, Android...) double check this thing and investigate how to avoid hard-coding static here

CMakeLists.txt
```diff
@@ -498,13 +544,23 @@ if(RS_JSON_API)
...
 			TIMEOUT 10
-#			EXCLUDE_FROM_ALL # Available only in CMake >= 3.28
+			# PATCH_COMMAND applies hotfixes to restbed before building:
+			# 1. Removes /wd4251 (MSVC-specific flag) that breaks GCC/Clang builds.
+			# 2. Bumps CMake minimum version to 3.5.0 to suppress deprecation warnings in CMake >= 3.27.
+			PATCH_COMMAND sed -i -e "s|/wd4251||g" -e "s|VERSION 3.1.0|VERSION 3.5.0|g" CMakeLists.txt
 		)
 		FetchContent_MakeAvailable(restbed)
 
+		if(TARGET restbed-shared)
```
G10h4ck: Static/dynamic linking hard-coding again, is this really necessary? Do it properly not hard-coding

CMakeLists.txt
```diff
@@ -619,7 +675,7 @@ if(RS_BRODCAST_DISCOVERY)
 endif(RS_BRODCAST_DISCOVERY)
 
 if(NOT RS_WARN_DEPRECATED)
-	target_compile_definitions(${PROJECT_NAME} PRIVATE RS_NO_WARN_DEPRECATED)
+	target_compile_definitions(${PROJECT_NAME} PUBLIC RS_NO_WARN_DEPRECATED)
```
G10h4ck: Why making it PUBLIC ?

CMakeLists.txt
```diff
@@ -735,7 +810,7 @@ if(RS_ANDROID AND CMAKE_BUILD_TYPE STREQUAL "Release")
 	set_property(TARGET ${PROJECT_NAME} PROPERTY VISIBILITY_INLINES_HIDDEN ON)
 endif()
 
-if(CMAKE_BUILD_TYPE STREQUAL "Release")
+if(CMAKE_BUILD_TYPE STREQUAL "Release" AND NOT WIN32)
```
G10h4ck, line 825: Disabling LTO will make RS slower and fatter in windows, but it is acceptable for now, add a comment explaining why is this necessary, and what the consequences are

G10h4ck (changes requested):
See my inline comments

src/pqi/sslfns.h
jolavillette: X509_NAME is actually not a RetroShare type; it is an internal OpenSSL struct. We cannot rename it to RS_X509_NAME because it is defined in OpenSSL's headers. The #undef is the only safe workaround to allow using OpenSSL's X509_NAME if Windows <wincrypt.h> (which defines it as a macro) was previously included.

src/pqi/sslfns.h
jolavillette: Same as above. X509_EXTENSIONS is an OpenSSL internal type, not a RetroShare type. We must undefine the Windows macro to be able to use OpenSSL properly.

src/retroshare/rstor.h
jolavillette: Good point. Undefining ERROR in a public header is dangerous for downstream users. I have removed the #undef ERROR hack entirely from rstor.h. Instead, I added WIN32_LEAN_AND_MEAN globally in CMakeLists.txt (as you suggested below), which completely prevents <wingdi.h> from polluting the namespace.

src/util/rsdebug.h
jolavillette: I have removed the #undef ERROR hack here as well, and I added target_compile_definitions(${PROJECT_NAME} PUBLIC WIN32_LEAN_AND_MEAN) in CMakeLists.txt for Windows builds. This cleanly solves the macro pollution issue without touching the headers.

CMakeLists.txt
jolavillette: I have replaced that hardcoded list with a simple -Wl,--exclude-libs,ALL. This cleaner approach prevents any statically linked library from blindly re-exporting its symbols on Windows, automatically handling any future dependency changes.

CMakeLists.txt
jolavillette: Thanks

CMakeLists.txt
jolavillette: You are right, hardcoding restbed-static was too restrictive and broke systems that don't install a system-wide librestbed.so. Simply reverting it to restbed caused ld: library 'restbed' not found locally because the fetched restbed project does not create an alias named restbed, so CMake was emitting -lrestbed instead of linking the local target. I have updated the logic to dynamically resolve the correct target (restbed-shared or restbed-static) based on BUILD_SHARED_LIBS, and link to that specific target using a ${RESTBED_TARGET} variable.

CMakeLists.txt
jolavillette: I have removed the hardcoded EXCLUDE_FROM_ALL and static/shared checks. We now rely strictly on the generic BUILD_SHARED_LIBS property and seamlessly link the resolved ${RESTBED_TARGET}.

CMakeLists.txt
jolavillette: That was a mistake. I have changed it to PRIVATE so we don't suppress deprecation warnings for downstream targets.

CMakeLists.txt
jolavillette: Done. I have added a clear comment explaining that LTO is disabled on Windows because compiling RetroShare with LTO enabled exhausts the MinGW ld memory and symbol limits due to the massive code size, which caused the build to fail.

jolavillette:
about librnp:
"libnrp does use CMake too? In that case it shouldn't be needed to compile it manually before, if added properly as a dependency CMake will deal with compiling of that too automatically just like the other dependencies (ex: restbed)"

You are completely right. We previously avoided adding it as a native CMake dependency because statically linking the librnp CMake target re-exported all of its internal dependencies (Botan, json-c, etc.) and broke the MinGW 65535 exported symbol limit on Windows. However, now that we have implemented the -Wl,--exclude-libs,ALL fix from your other comment, this limitation is gone. We can absolutely integrate librnp properly via CMake

src/pqi/sslfns.h
G10h4ck: double check if this still happens with WIN32_LEAN_AND_MEAN

src/pqi/sslfns.h
G10h4ck: double check if this still happens with WIN32_LEAN_AND_MEAN

src/pqi/sslfns.h
jolavillette: Done. I removed the #undef workaround and rely on WIN32_LEAN_AND_MEAN instead, which keeps <windows.h> from pulling in <wincrypt.h> (where the colliding X509_NAME macro comes from). It was already defined globally for the qmake build and for the CMake shared target, but not for the CMake static target — which is the default (RS_LIBRETROSHARE_STATIC=ON) — so I moved the definition out of the RS_LIBRETROSHARE_SHARED block to cover both. The only explicit #include <wincrypt.h> is in rsserver/rsloginhandler.cc (for DPAPI), and that file doesn't include sslfns.h, so no translation unit ever sees both. I also fixed the misleading comment: it's an OpenSSL type, not a RetroShare one.
Compilation OK linux, windows, macos

src/pqi/sslfns.h
jolavillette: Same as above — removed together with the X509_NAME workaround. With WIN32_LEAN_AND_MEAN now applied to the default (static) CMake target as well, <wincrypt.h> is never included in a translation unit that uses the OpenSSL types, so the X509_EXTENSIONS macro collision can no longer occur.

CMakeLists.txt
G10h4ck: ping @jolavillette

CMakeLists.txt
G10h4ck: ping @jolavillette

CMakeLists.txt
G10h4ck: Just to make sure, double check if this is still needed after the excluding of used libraries symbols exporting

G10h4ck (changes requested):

jolavillette:
1. These definitions must be PUBLIC (not PRIVATE) because downstream targets (retroshare-gui, retroshare-service) include libretroshare's public headers, which in turn include <windows.h> and <openssl/x509.h>. If these flags are only PRIVATE, the macro conflicts reappear in every downstream target that includes our headers.


2. Each of these solves a specific, documented compilation failure on MSYS2/MinGW:

UNICODE / _UNICODE: libbitdht/src/util/bdfile.cc:47 calls MoveFileExW() with wchar_t* arguments. Without UNICODE, the Windows headers resolve the call to MoveFileExA() which expects char*, producing: error: cannot convert 'const wchar_t*' to 'LPCSTR'.

NOGDI: <windows.h> includes <wingdi.h> which does #define ERROR 0. This macro silently replaces every occurrence of ERROR in RetroShare's logging code (e.g. RsLog levels in rsdebug.h). Note: WIN32_LEAN_AND_MEAN does not exclude <wingdi.h>, so NOGDI is needed explicitly. This replaces the #undef ERROR hacks that were previously in rstor.h and rsdebug.h (removed per your earlier review comments #4 and #5).

NOMINMAX: <windows.h> defines #define min(a,b) and #define max(a,b) macros that break every call to std::min() / std::max() throughout the codebase.

NOCRYPT: <windows.h> includes <wincrypt.h> which does #define X509_NAME ((LPCSTR) 7) and #define X509_EXTENSIONS ((LPCSTR) 5). These macros shadow OpenSSL's X509_NAME and X509_EXTENSIONS structs, making any #include <openssl/x509.h> fail. This is the clean replacement for the #undef workaround that was in sslfns.h (removed per your review comments #2 and #3).


3. --exclude-libs and LTO are independent, and disabling IPO/LTO on Windows is still required: the build does not link with LTO enabled on Windows (MinGW/GCC LTO codegen fails with multiple-definition / thunk errors), regardless of the symbol exclusion.

The exclusion you're referring to is a separate fix, on the shared/DLL build only: it had to be narrowed from --exclude-libs,ALL (which broke the build) down to just the heavy third-party archives (libbitdht, librnp, libsexpp, libudp-discovery, librestbed, libsqlcipher, libbotan-2/3, libz, libbz2) — enough to stay under the PE/COFF ~65535 exported-symbol cap while keeping libretroshare's own API exported. The default static build is unaffected.

So AND NOT WIN32 should stay.

G10h4ck:
@jolavillette please keep replying comment by comment, I have very little time, so try to optimize for the reviewer time ;-)

CMakeLists.txt
```diff
@@ -227,8 +232,24 @@ if(RS_LIBRETROSHARE_SHARED)
 	## Ensure statically linked libraries such as openpgpsdk are compiled with
 	## PIC Which is needed for shared library
 	set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+
+	if(WIN32)
+		set_target_properties(${PROJECT_NAME} PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)
+		if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+			target_link_options(${PROJECT_NAME} PRIVATE "-Wl,--export-all-symbols,--exclude-libs,libbitdht.a:librnp.a:libsexpp.a:libudp-discovery.a:librestbed.a:libsqlcipher.a:libbotan-2.a:libbotan-3.a:libz.a:libbz2.a")
```
G10h4ck, line 242: Isn't any way to do this without listing all the libraries on windows, like `ALL` on linux? Also is `if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")` needed, and why?

CMakeLists.txt
```diff
@@ -227,8 +232,24 @@ if(RS_LIBRETROSHARE_SHARED)
...
 endif(RS_LIBRETROSHARE_SHARED)
 
+if(WIN32)
+	# Define WIN32_LEAN_AND_MEAN for both the static and shared library targets
+	# to prevent <windows.h> from including extra headers such as <wingdi.h>
+	# (which pollutes the global namespace with macros like ERROR) and
+	# <wincrypt.h> (whose X509_NAME / X509_EXTENSIONS macros collide with the
+	# OpenSSL types of the same name used in pqi/sslfns.h).
+	target_compile_definitions(${PROJECT_NAME} PUBLIC WIN32_LEAN_AND_MEAN)
+endif()
```
G10h4ck: I think `WIN32_LEAN_AND_MEAN` should go at line 278 or so, this also would avoid another `if` block and code lines to maintain

CMakeLists.txt
G10h4ck: Also please add a comment in the code explaining why those defines are needed, if necessary each of them @jolavillette

CMakeLists.txt
G10h4ck: ping @jolavillette

G10h4ck (changes requested):

CMakeLists.txt
jolavillette: To answer your first question: unfortunately, no. I just tested --exclude-libs,ALL on Windows, and the linkage of downstream targets (like retroshare-gui) fails completely with massive undefined reference errors to libretroshare's own core symbols (e.g., ChatId, RsTypeSerializer, etc.).

The reason is a quirk in how CMake implements WINDOWS_EXPORT_ALL_SYMBOLS ON. During the build, CMake bundles all the compiled source files (.obj) into a temporary internal archive (e.g., CMakeFiles/retroshare.dir/objects.a) before passing it to the linker. If we use --exclude-libs,ALL, the GNU linker aggressively hides symbols from every archive it encounters, including CMake's own internal objects.a. Consequently, libretroshare's own C++ symbols are stripped from the exports. The explicit list is the only way to hide the massive 3rd-party dependencies (avoiding the 65535 limit) while allowing CMake's internal archive to safely export our own code.

To answer your second question: yes, the if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang") check is strictly needed. The -Wl,... syntax is exclusively understood by GNU/Clang linkers (ld or lld). If a developer configures the Windows build using MSVC (cl.exe / link.exe), passing -Wl,... flags will cause a fatal linker error. This check safely restricts the flag to compatible MinGW/Clang environments.

G10h4ck: In that case please add a comment explaining shortly this situation, always remember if someone else read that code 3 years from now they should be able to understand the choice

CMakeLists.txt
G10h4ck: ping @jolavillette

G10h4ck (changes requested):

CMakeLists.txt
jolavillette: I don't know why I can reply here and not under the next comments...

1. I have moved WIN32_LEAN_AND_MEAN to the main if(WIN32) block alongside the other Windows definitions, and removed the redundant if(WIN32) block entirely.

2. "Done. I have added inline comments right above each definition inside the CMakeLists.txt file explaining exactly what header pollution or compilation error they prevent (MoveFileExW translation, <wingdi.h> ERROR macro collision, min/max collision, and OpenSSL X509 types shadowing).

3. You're right, that find_library(BOTAN_LIBRARY ...) line was dead code leftover from before we introduced add_subdirectory(librnp). Now that librnp handles its own dependencies internally through CMake, we don't need to manually find Botan in libretroshare's CMakeLists anymore. I have completely removed it.

4.  I have just added a detailed comment above the --exclude-libs line explaining that we use an explicit list instead of ALL because ALL inadvertently hides symbols from CMake's internal objects.a archive, preventing libretroshare from exporting its own core C++ symbols.

5. Regarding WIN32_LEAN_AND_MEAN and the inline comments for the definitions, this was already addressed and fixed in my latest commit. The flags are now cleanly grouped together with proper comments for each of them.

G10h4ck (approved):
Legit

PR #308, opened 2026-06-08
fix issues in gxstrans

jolavillette:
1. ACK stamp - processRecvdMessages now advertises messages injected via receiveNewMessages (the GxsTrans receipt) to friends; without it the ACK never propagates back.
2. notifyGxsTransSendStatus - break inside the if (x2): the linear search over msgOutgoing no longer bails out after the 1st bucket.
3. notifyGxsTransSendStatus - it->first instead of msg_id (x2): the GUI event references the original message, not its per-recipient copy.
4. notifyDataStatus - found flag instead of the premature return: same linear-search bug.
5. checkOutgoingMessages - emits a change notification when the PENDING flag is cleared (GUI refresh).

### DIFF
--- a/src/gxs/rsgenexchange.cc
+++ b/src/gxs/rsgenexchange.cc
@@ -3094,6 +3094,7 @@ void RsGenExchange::computeHash(const RsTlvBinaryData& data, RsFileHash& hash)
 void RsGenExchange::processRecvdMessages()
 {
     std::list<RsGxsMessageId> messages_to_reject ;
+    std::set<RsGxsGroupId> grps_with_new_msgs ;	// groups that received new messages, to stamp their server update TS off-mutex below
 
     {
 	    RS_STACK_MUTEX(mGenMtx) ;
@@ -3273,6 +3274,12 @@ void RsGenExchange::processRecvdMessages()
 
             for(auto& nxs_msg: msgs_to_store)
             {
+                // msgs_to_store is now post-deduplication (removeDeleteExistingMessages above), so this only
+                // contains genuinely new messages. Mark their groups for the server-TS stamp here, NOT from
+                // groups_last_post_update (which is populated pre-dedup and would also fire for already-known
+                // messages re-received via sync, causing redundant stamps/advertisements).
+                grps_with_new_msgs.insert(nxs_msg->grpId);
+
                 RsGxsMsgItem *item = dynamic_cast<RsGxsMsgItem*>(mSerialiser->deserialise(nxs_msg->msg.bin_data,&nxs_msg->msg.bin_len));
 
                 if(!item)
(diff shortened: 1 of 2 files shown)


src/gxstrans/p3gxstrans.cc
```diff
@@ -866,6 +873,26 @@ bool p3GxsTrans::dispatchDecryptedMail( const RsGxsId& authorId,
 	          << "with: msgId: " << receipt->msgId << std::endl;
 #endif
 
+	RsGxsTransId mailId = 0;
```
csoler: This entire block is doing nothing except printing dbg info. Please use a ifdef around.

src/grouter/p3grouter.cc
```diff
@@ -1074,6 +1070,8 @@ void p3GRouter::routePendingObjects()
 
 void p3GRouter::locked_collectAvailableFriends(const GRouterKeyId& gxs_id,const std::set<RsPeerId>& incoming_routes,uint32_t duplication_factor, std::map<RsPeerId,uint32_t>& friend_peers_and_duplication_factors)
 {
+    return; // FORCE NO DIRECT FRIENDS AVAILABLE TO TEST TUNNEL FALLBACK ROUTING!
```
csoler: I think you do not want to keep that debugging hack.

src/grouter/p3grouter.cc
```diff
@@ -1287,6 +1286,25 @@ bool p3GRouter::locked_sendTransactionData(const RsPeerId& pid,const RsGRouterTr
 {
     if(mTurtle->isTurtlePeer(pid))
     {
+        if (trans_item.PacketSubType() == RS_PKT_SUBTYPE_GROUTER_TRANSACTION_CHUNK)
```
csoler: same, this blobk is for debug only and is fairly large. If temporary, jusr remove it.

src/grouter/p3grouter.cc
```diff
@@ -1318,6 +1336,25 @@ bool p3GRouter::locked_sendTransactionData(const RsPeerId& pid,const RsGRouterTr
     }
     else
     {
+        if (trans_item.PacketSubType() == RS_PKT_SUBTYPE_GROUTER_TRANSACTION_CHUNK)
```
csoler: same

src/grouter/p3grouter.cc
```diff
@@ -1360,6 +1397,17 @@ void p3GRouter::autoWash()
                     
             if(delete_entry)
             {
+                if( it->second.data_status != RS_GROUTER_DATA_STATUS_DONE && it->second.data_status != RS_GROUTER_DATA_STATUS_RECEIPT_OK )
```
csoler: same. I won't comment on the remaining ones.

src/gxs/rsgenexchange.cc
```diff
@@ -3299,14 +3301,30 @@ void RsGenExchange::processRecvdMessages()
...
     }
 
     // Done off-mutex to avoid cross deadlocks in the netservice that might call the RsGenExchange as an observer..
 
     if(mNetService != NULL)
+    {
 	    for(std::list<RsGxsMessageId>::const_iterator it(messages_to_reject.begin());it!=messages_to_reject.end();++it)
 		    mNetService->rejectMessage(*it) ;
+
+	    // MAIL: stamp the server-side msg update TS for groups that received new messages, so that friends
```
csoler: We're in rsgenexchange here so there's nothing specific to each service. Update the comment.

src/gxs/rsgenexchange.cc
```diff
@@ -3299,14 +3301,30 @@ void RsGenExchange::processRecvdMessages()
...
+	    // get notified and re-synchronise. For messages received through the regular netservice transaction
+	    // path this is already done in RsGxsNetService::processCompletedTransactions(), but messages injected
+	    // directly via receiveNewMessages() (e.g. GxsTrans presigned receipts / ACKs) bypass that path and
+	    // would otherwise be stored but never advertised to friends.
+	    for(const RsGxsGroupId& grpId : grps_with_new_msgs)
+	    {
+		    RsDbg() << "MAIL (" << AuthSSL::getAuthSSL()->getOwnLocation() << "): GXS - Stamping server msg update TS for group " << grpId
+		            << " after injecting new message(s) via receiveNewMessages() (e.g. GxsTrans ACK), so friends get notified and re-sync";
+		    mNetService->stampMsgServerUpdateTS(grpId) ;
```
csoler, line 3325: ok this is probably where the problem happens. Good catch.

csoler (changes requested):
- for debugging purpose you added a log of RsDbg(). That's fine, but you should use #ifdefs around (GROUTER_DEBUG, GXSTRANS_DEBUG, etc), so we can easily activate them service by service. For most debug blocks (some are large!) that actually helped you find the problem, just remove them.

PR #310, opened 2026-06-16
build(cmake): describe the super-project for the displayed version

jolavillette:
ba90b271 dropped the parent-repo detection in the version block, so RS_MAJOR/MINOR/MINI/EXTRA_VERSION (and thus the GUI "RetroShare Version") now describe the libretroshare submodule instead of the RetroShare super-project. Restore the ../.git detection: describe the parent repo when libretroshare is built as a submodule, fall back to this directory for standalone / FetchContent / Android builds.


G10h4ck:
This is just wrong, @jolavillette do you remember what we discussed in the chat? RetroShare-Gui needs its own version macros indipendent from libretroshare. Then RetroShare-Gui will display both its versioning information, and libretroshare information as it does with other libraries! And this broken Android build again.

PR #314, opened 2026-06-26
Chat: fix false time-shift alerts by moving statistics after deduplcation

jolavillette:
Chat: fix false time-shift alerts by moving statistics after deduplcation

Problem
During sustained activity in a chat lobby, RetroShare would wrongly raise the TIME_SHIFT_PROBLEM event (a warning that a friend's system clock is out of sync), even when no clock was actually wrong.

Root cause
In a lobby, the same message is relayed by several friends, so we receive N identical copies. Deduplication is handled by bounceLobbyObject(), which returns false when the msg_id is already in the cache (an echo).

However, addTimeShiftStatistics() was being called before that deduplication (and even before the "message too old / from the future" checks). As a result, a single copy of a message carrying a questionable sendTime was counted once per relay. The time-shift histogram filled up with duplicates of the same event, the median crossed the alert threshold, and a false positive was emitted.

The existing "at most 1 sample per second" rate limiter wasn't enough, because the duplicate copies typically arrive in a burst within the same second.

Fix
Moved the addTimeShiftStatistics() call to after bounceLobbyObject() in handleRecvChatLobbyEventItem(). Now only a genuinely new message (not a duplicate, not stale, not from the future) contributes a statistic → one message = one sample, regardless of how many relays delivered it.
Changed the signature to addTimeShiftStatistics(int shift, const RsGxsId& gxsId) so each sample can be attributed to the sending identity.

Diagnostics (logging)
To make any future real clock drift easier to investigate, the function now logs via RsDbg():
The debug outputs are guarded by #ifdef DEBUG_CHAT_LOBBIES
[TS-ACCEPT] — each accepted sample (identity, delay, log bin).
[TS-STATS] — at every 30-vote window, the estimated median delay and the list of contributors (nickname(count)).
[TS-ALERT] — when TIME_SHIFT_PROBLEM is actually emitted.
The computation itself (log2 histogram, interpolated median, expected > 9 threshold) is unchanged; only the timing of the call and the instrumentation were added.

### DIFF
--- a/src/chat/distributedchat.cc
+++ b/src/chat/distributedchat.cc
@@ -609,72 +609,94 @@ void DistributedChatService::handleRecvChatLobbyList(RsChatLobbyListItem *item)
     _should_reset_lobby_counts = false ;
 }
 
-void DistributedChatService::addTimeShiftStatistics(int D)
+void DistributedChatService::addTimeShiftStatistics(int D, const RsGxsId& gxsId)
 {
-	// Bursts of messages from friends using a wrong system clock can trigger a TIME_SHIFT_PROBLEM event 
-	// We eliminate that by taking into account at most 1 message per second
-	static rstime_t last_stat_time = 0;
-	rstime_t now = time(NULL);
-	if(now <= last_stat_time)
-		return;
-	last_stat_time = now;
+    // --- GLOBAL RATE LIMITER ---
+    // We only take at most 1 sample per second from the whole network.
+    // This provides robustness against network bursts/batching from relays.
+    static rstime_t last_stat_time = 0;
+    rstime_t now = time(NULL);
 
-	static const int S = 50 ; // accuracy up to 2^50 second. Quite conservative!
-	static int total = 0 ;
-	static std::vector<int> log_delay_histogram(S,0) ;
+    if (now <= last_stat_time) {
+        return;
+    }
+    last_stat_time = now;
 
-	int delay = (D<0)?(-D):D ;
+    // Resolve identity name for readable logs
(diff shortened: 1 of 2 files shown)


src/chat/distributedchat.cc
```diff
@@ -609,72 +609,104 @@ void DistributedChatService::handleRecvChatLobbyList(RsChatLobbyListItem *item)
...
-	int bin = std::min(S-1,l) ;
-	++log_delay_histogram[bin] ;
-	++total ;
+    int bin = std::min(S - 1, l);
+    ++log_delay_histogram[bin];
+    ++total;
 
 #ifdef DEBUG_CHAT_LOBBIES
-	std::cerr << "New delay stat item. delay=" << D << ", log=" << bin << " total=" << total << ", histogram = " ;
+    contributors[gxsId]++;
```
csoler: I don't think that the int is 0 on default on every possible g++ implementations.

csoler (changes requested):
This change seems appropriate to me.
Make sure about the initialization of std::map<something,int>, when you do mymap[something]++;
Also if possible, ask the AI to not re-write parts of the code (adding spaces, {}s, etc) unless required by the changes it's doing (it draws unnecessary attention).

src/chat/distributedchat.cc
jolavillette: I've changed it to insert the counter with an explicit 0 (contributors.insert(std::make_pair(gxsId,0)).first->second++) so it no longer relies on value-initialization. (This is debug-only code, guarded by DEBUG_TIME_SHIFT.)

PR #317, opened 2026-07-03
build(cmake): add options to build the GXS Wiki and The Wire services

jolavillette:
The CMake build never compiled the experimental GXS "Wiki" (WikiPoos) and "The Wire" services: their sources were commented out in src/CMakeLists.txt and no option defined the RS_USE_WIKI / RS_USE_WIRE macros that gate the service instantiation in rsserver/rsinit.cc and the matching GUI code. As a result the retroshare-gui RS_GXSWIKIPOS / RS_GXSTHEWIRE blocks could not be enabled to a working build.

Mirror the qmake "wikipoos" / "gxsthewire" CONFIG switches:

- Add RS_GXSWIKIPOS and RS_GXSTHEWIRE options (default OFF, as in qmake).
- Compile the p3wiki/p3wire services, rswikiitems/rswireitems and the rswiki.h/rswire.h public headers when the respective option is enabled.
- Define RS_USE_WIKI / RS_USE_WIRE PUBLIC so libretroshare consumers see the same guard.

Both features stay OFF by default; they are unmaintained and experimental.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -364,9 +364,6 @@ list(
 	rsitems/rsrttitems.cc
 	rsitems/rsserviceinfoitems.cc )
 
-# NOTE: rswikiitems / rswireitems sources are added further below in the
-# services section, guarded by the RS_GXSWIKIPOS / RS_GXSTHEWIRE options.
-
 list(
 	APPEND RS_SOURCES
 	rsitems/rsgxschannelitems.cc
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -363,10 +363,8 @@ list(
 	rsitems/rsrttitems.cc
 	rsitems/rsserviceinfoitems.cc )
 
-#retroshare/rswiki.h
-#./rsitems/rswikiitems.cc
-#./rsitems/rswikiitems.h
-#./rsitems/rswireitems.h
+# NOTE: rswikiitems / rswireitems sources are added further below in the
```
csoler: plz remove this comment. It's not needed anymore since the removed elements are...removed!

jolavillette:
done

PR #322, opened 2026-07-05
fix(peermgr): default SSL-only friends to RS_NODE_PERM_NONE

jolavillette:
fix(peermgr): default SSL-only friends to RS_NODE_PERM_NONE

addSslOnlyFriend() registers a friend without creating an entry in mFriendsPermissionFlags. This is the path taken by every short-invite add (webui, mobile, QR code). Later, gossip discovery (p3gossipdiscovery.cc) and loadList() call addFriend() with the RS_NODE_PERM_ALL "keep existing flags" mask, which is only reduced (service_flags &= entry) when a per-friend entry already exists.

For SSL-only friends no entry exists, so the RS_NODE_PERM_ALL mask is written verbatim and persisted, silently granting the peer DIRECT_DL | ALLOW_PUSH | REQUIRE_WL. The stray REQUIRE_WL then makes incoming connections from non-whitelisted addresses get refused (NOT_WHITELISTED) in pqissl::accept_locked(), spamming the log with "This should never happen at this point!" stack traces.

GUI-added friends are unaffected because addFriend() always writes an entry (RS_NODE_PERM_DEFAULT). Establish the same default here so the downstream mask stays a no-op.

### DIFF
--- a/src/pqi/p3peermgr.cc
+++ b/src/pqi/p3peermgr.cc
@@ -1185,6 +1185,19 @@ bool p3PeerMgrIMPL::addSslOnlyFriend( const RsPeerId& sslId, const RsPgpId& pgp_
 	{ RS_STACK_MUTEX(mPeerMtx);
 		mFriendList[sslId] = pstate;
 		mStatusChanged = true;
+
+		/* Establish default (empty) service permissions for this PGP id if none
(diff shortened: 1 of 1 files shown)


csoler:
Nice find. For the record, that SSL only friend system is a huge security hole. I'm all for getting rid of it as soon as we have ECC 25519 keys.

PR #323, opened 2026-07-05
fix(events): don't warn/stacktrace on a valid pre-generated handler id

jolavillette:
fix(events): don't warn/stacktrace on a valid pre-generated handler id

registerEventsHandler() documents (rsevents.h) that the caller may pass an hId previously obtained from generateUniqueHandlerId(). The JSON API event-stream wrapper (jsonapi.cpp) does exactly this: it pre-generates the id so its SSE callbacks can capture it to unregister themselves, then passes it to registerEventsHandler().

The implementation treated any non-zero hId as an override, calling print_stacktrace() + "Overriding an existing error handler ID ... very unexpected". So every event-stream registration (every webui/mobile connection, plus each 5s reconnect and each 1MB queue restart) dumped a bogus stack trace to the log, contradicting the API's own contract.

Only warn when the hId is actually already present in the handler maps, i.e. a real override. A freshly generated id never is (ids are monotonic), so the false positive is gone while genuine accidental id reuse is still reported. Only jsonapi.cpp pre-generates ids; every other caller passes 0, so behaviour there is unchanged.

### DIFF
--- a/src/services/rseventsservice.cc
+++ b/src/services/rseventsservice.cc
@@ -138,8 +138,25 @@ std::error_condition RsEventsService::registerEventsHandler(
         hId = generateUniqueHandlerId_unlocked();
     else
     {
-        print_stacktrace();
-        RsWarn() << "Overriding an existing error handler ID with a new callback. This is very unexpected. Make sure you know what you are doing." ;
+        /* A non-zero hId is a legitimate, documented use case: the caller may
+         * provide an id previously obtained from generateUniqueHandlerId() (see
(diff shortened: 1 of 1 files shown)


csoler:
Very surprising PR: the check was added because it's very easy (due to a design flaw) to not initialize a handle ID and replace an existing one by mistake. Normally generateUniqueHandlerId() should be private so that it is only called by the  register method itself. Besides, if you allow to bypass the call to generateUniqueHandlerId(), then mLastHandler will not be consistent with the existing list of Ids anymore.

PR #330, opened 2026-07-11
RsEventsService: make unregisterEventsHandler() a barrier (fix shutdown UAF)

jolavillette:
RsEventsService: make unregisterEventsHandler() a barrier (fix shutdown UAF)

handleEvent() copies the registered callbacks under mHandlerMapMtx and then runs them *outside* the lock (on purpose, so a callback may re-enter to send events or unregister itself). As a side effect unregisterEventsHandler() was not a barrier: it removed the handler from the map, but an in-flight dispatch still held a private copy of it. If the handler's owner (a GUI widget) was destroyed on another thread in that window, the copy fired against a dangling object -> use-after-free, typically SIGSEGV in qobject_cast<QThread*> inside RsQThreadUtils::postToObject. This is what tears the process down at shutdown, when many widgets are destroyed while the events thread keeps ticking a flood of peer-disconnect events (the rest of the crash dump is just exit()-in-the- signal-handler unwinding RsServer while other threads are still running).

Add a recursive mDispatchMtx held by handleEvent() across the whole dispatch loop. unregisterEventsHandler() erases the handler under mHandlerMapMtx (no future dispatch can pick it up) and then fences on mDispatchMtx (no ongoing dispatch is still running it), so once it returns the handler is guaranteed neither running nor about to start. Widgets that unregister from their destructor are then safe to destroy. Recursive so a callback re-entering on the dispatching thread (self-unregister / synchronous sendEvent) does not deadlock; safe because handlers post asynchronously (Qt QueuedConnection) and never block on the unregistering thread, so there is no lock-order cycle.

### DIFF
--- a/src/services/rseventsservice.cc
+++ b/src/services/rseventsservice.cc
@@ -166,18 +166,33 @@ std::error_condition RsEventsService::registerEventsHandler(
 std::error_condition RsEventsService::unregisterEventsHandler(
         RsEventsHandlerId_t hId )
 {
-	RS_STACK_MUTEX(mHandlerMapMtx);
+	std::error_condition retval = RsEventsErrorNum::INVALID_HANDLER_ID;
 
-	for(uint32_t i=0; i<mHandlerMaps.size(); ++i)
 	{
-		auto it = mHandlerMaps[i].find(hId);
-		if(it != mHandlerMaps[i].end())
+		RS_STACK_MUTEX(mHandlerMapMtx);
+
+		for(uint32_t i=0; i<mHandlerMaps.size(); ++i)
 		{
-			mHandlerMaps[i].erase(it);
-			return std::error_condition();
+			auto it = mHandlerMaps[i].find(hId);
+			if(it != mHandlerMaps[i].end())
+			{
+				mHandlerMaps[i].erase(it);
+				retval = std::error_condition();
+				break;
+			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


G10h4ck:
@jolavillette the commit seems to make sense, I have some questions

1) the error you are experiencing is reproducible?
2) do you have proofread the changes?
3) the patched version does consistently work without the error you where experiencing before?
4) A human pass over the comments to avoid redundance and LLM verbosity would be welcome

jolavillette:
1. Not easily reproductible
The crash was reported by Opax, with a console log only, not gdb log

2 3 and 4. I now think that I should not have submitted the console log to the AI, and asked for an explanation and a fix. So let's just ignore this PR for master, after all it is just a crash on exit. I won't do that mistake again.

There are several more important bugfixing PRs waiting for reviews since weeks. 

Among them are 6 PRs that are needed to make cmake build work on Linux/macOS/Win Qt5/Qt6, and they are still not merged:  3242 3243 3244 3250 3269 and 3270.

csoler:
It seems to make sense, although it took me a while to figure out how that actually works (the very verbose explanations of the AI are not helping): as far as I understand, the point here is that when a QObject unregisters a callback in its destructor, it will have to wait on mDispatchMtx lock (inside the unregister call) as long as an event dispatch is running (whichever the destination). Putting the lock in the QObject itself (in handleEvent_mainThread() and destructor()), is another possibility but it would need to patch each widget registering a rsEvent.

PR #331, opened 2026-07-12
Fix UI freeze on "mark all as read/unread" for forums, channels and boards

jolavillette:
works with retroshare-gui pr/3274

Fix UI freeze on "mark all as read/unread" for forums, channels and boards

## Problem

Marking a large GXS group read — right-click → **Mark all as read** (or **unread**) — froze the UI for a very long time and could appear to hang indefinitely. On an 8000-post forum the hourglass stayed up for over an hour; force-quitting mid-way left thousands of posts still unread.

## Root cause

Two independent amplifiers, both triggered *per message*:

1. **GUI side** — the model spawned **one detached `std::thread` per changed post** (each blocking up to 5s waiting for its token to complete) and emitted **one `dataChanged()` per post**. Marking N posts created ~N threads and ~N signals, all driven from the GUI thread.
2. **Backend side** — `RsGenExchange::processMsgMetaChanges()` wrote each status change to the GXS database in **its own implicit transaction** (one fsync per message). On the SQLCipher-encrypted store this meant thousands of serial encrypted commits.

## Fix

**Backend (`libretroshare`)**

- `RsGeneralDataService`: new batch `updateMessageMetaData(std::vector<MsgLocMetaData>)` overload (default loops over the single version). `RsDataService` overrides it to persist the whole batch in **a single transaction**, held under one DB mutex so nothing interleaves.
- `RsGenExchange::processMsgMetaChanges()`: resolve all status masks first, then persist every queued change through that single transaction. This benefits **every** GXS service.
- New bulk read-status APIs that queue all changes at once and emit a **single** event:
  - `RsGxsForums::markRead(forumId, msgIds, read)`
  - `RsGxsChannels::setMessageReadStatus(channelId, msgIds, read)`
  - `RsPosted::setPostReadStatus(boardId, msgIds, read)`

**GUI (`retroshare-gui`)**

- `GxsForumModel`, `GxsChannelPostsModel` and `PostedPostsModel`: "mark all as read/unread" now collects the affected message ids, updates the in-memory model, issues **one** background batch call, and refreshes the view **once** — instead of one detached thread + one event + one `dataChanged()` per post.

## Result

The GUI thread now does `O(n)` in-memory work plus **one** thread and **one** signal, regardless of how many posts are affected, so it stays responsive on arbitrarily large groups. Persistence happens in a **single transaction** (one fsync). Works symmetrically for **read and unread**, on **forums, channels and boards**.

## Notes

- Backwards compatible: the existing single-message `markRead` / `setMessageReadStatus` / `setPostReadStatus` are unchanged; the batch variants are additive.
- This PR pairs with the companion in the other repository (`retroshare-gui` ⇄ `libretroshare`): the GUI batch calls require the new backend overloads. Companion PR: https://github.com/RetroShare/RetroShare/pull/3274

## Testing

Open a forum/channel/board with thousands of posts, then right-click → **Mark all as read**, and again → **Mark all as unread**:

- the UI stays responsive throughout (no hourglass hang);
- unread counts update immediately;
- the state persists across a restart.

### DIFF
--- a/src/gxs/rsgds.h
+++ b/src/gxs/rsgds.h
@@ -25,6 +25,7 @@
 #include <set>
 #include <map>
 #include <string>
+#include <vector>
 
 #include "inttypes.h"
 
@@ -247,6 +248,25 @@ public:
      * @param metaData
      */
     virtual int updateMessageMetaData(const MsgLocMetaData& metaData) = 0;
+
+    /*!
+     * @brief Update the meta data of several messages at once.
+     *
+     * The default implementation just updates them one by one. Stores backed by
+     * a transactional database should override this to persist the whole batch
+     * in a single transaction, which is dramatically faster when marking
+     * thousands of messages (e.g. "mark all as read" on a large forum).
+     *
+     * @param metaList the meta data items to update
+     * @return the number of items successfully updated
+     */
+    virtual int updateMessageMetaData(const std::vector<MsgLocMetaData>& metaList)
+    {
+        int count = 0;
+        for(const MsgLocMetaData& m : metaList)
+            count += updateMessageMetaData(m);
+        return count;
+    }
 
     /*!
      * @param metaData
--- a/src/gxs/rsgds.h
+++ b/src/gxs/rsgds.h
@@ -252,21 +252,14 @@ public:
     /*!
      * @brief Update the meta data of several messages at once.
      *
-     * The default implementation just updates them one by one. Stores backed by
-     * a transactional database should override this to persist the whole batch
+     * Stores backed by a transactional database should persist the whole batch
      * in a single transaction, which is dramatically faster when marking
      * thousands of messages (e.g. "mark all as read" on a large forum).
      *
      * @param metaList the meta data items to update
      * @return the number of items successfully updated
      */
-    virtual int updateMessageMetaData(const std::vector<MsgLocMetaData>& metaList)
-    {
-        int count = 0;
-        for(const MsgLocMetaData& m : metaList)
-            count += updateMessageMetaData(m);
-        return count;
-    }
+    virtual int updateMessageMetaData(const std::vector<MsgLocMetaData>& metaList) = 0;
 
     /*!
      * @param metaData
--- a/src/services/p3gxschannels.cc
+++ b/src/services/p3gxschannels.cc
@@ -2041,6 +2041,46 @@ bool p3GxsChannels::setMessageReadStatus(const RsGxsGrpMsgIdPair &msgId, bool re
 	return true;
 }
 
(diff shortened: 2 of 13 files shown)


src/gxs/rsgds.h
```diff
@@ -248,6 +249,25 @@ class RsGeneralDataService
      */
     virtual int updateMessageMetaData(const MsgLocMetaData& metaData) = 0;
 
+    /*!
```
csoler, line 252: I don't see the need for a default implementation here. The GUI PR is not supposed to work without the libretrohare part anyway. I think you can safely remove this one and leave the method pure virtual as updateMessageMetaData() above is. Also putting code into a .h is not so nice either.

src/services/p3gxschannels.cc
```diff
@@ -2041,6 +2041,46 @@ bool p3GxsChannels::setMessageReadStatus(const RsGxsGrpMsgIdPair &msgId, bool re
...
+
+	for(const RsGxsMessageId& msgId : msgIds)
+	{
+		uint32_t token;
+		setMsgStatusFlags(token, RsGxsGrpMsgIdPair(channelId, msgId), status, mask);
+		tokens.push_back(token);
+	}
+
+	// All tokens complete in the same tick, so waiting on the last is enough.
+	waitToken(tokens.back(), std::chrono::milliseconds(30000));
```
csoler: this optimization is based on assumptions about the implementation the setMsgStatusFlags function. I think it's dangerous to assume that the actions will be taken in the same order (suppose we decide to parallelize them). A loop over all tokens will not take more time but it certainly more correct.

src/services/p3gxsforums.cc
```diff
@@ -895,6 +895,49 @@ bool p3GxsForums::markRead(const RsGxsGrpMsgIdPair& msgId, bool read)
...
+	tokens.reserve(msgIds.size());
+
+	for(const RsGxsMessageId& msgId : msgIds)
+	{
+		uint32_t token;
+		setMsgStatusFlags(token, RsGxsGrpMsgIdPair(forumId, msgId), status, mask);
+		tokens.push_back(token);
+	}
+
+	// Wait for the whole batch to be processed. All tokens complete in the same
```
csoler: same here.

csoler (changes requested):

src/gxs/rsgds.h
jolavillette: done — pure virtual now.

src/services/p3gxschannels.cc
jolavillette: done — now waiting on every token. Same fix applied to p3posted.cc which had the same pattern.

src/services/p3gxsforums.cc
jolavillette: done.

PR #332, opened 2026-07-15
Added auth token fix for rs mobile

defnax:


src/retroshare/rsinit.h
```diff
@@ -419,7 +419,7 @@ class RsLoginHelper
 
 	/**
 	 * @brief Normal way to attempt login
-	 * @jsonapi{development,unauthenticated}
+	 * @jsonapi{development,manualwrapper}
 	 * @param[in] account Id of the account to which attempt login
 	 * @param[in] password Password for the given account
```
G10h4ck, line 424: Missing documentation for additional JSON API only params, look at createLocationV2 to see how to document them.

G10h4ck (changes requested):
See my inline comments

src/libretroshare.pro
```diff
@@ -940,7 +940,7 @@ rs_jsonapi {
     }
 
     DOXIGEN_INPUT_DIRECTORY=$$clean_path($${PWD})
-    DOXIGEN_CONFIG_SRC=$$clean_path($${RS_SRC_PATH}/jsonapi-generator/src/jsonapi-generator-doxygen.conf)
+    DOXIGEN_CONFIG_SRC=$$clean_path($${PWD}/jsonapi/jsonapi-generator-doxygen.conf)
```
G10h4ck: This is unrelated to this PR, and most probably would break on other platform, please discard this change. Also please do no do any mroe development in qmake files, they are temporarily kept meanwhile all distributions build scripts get ported to CMake. So all the attention should now move to CMake build system on windows too.

G10h4ck (changes requested):

src/libretroshare.pro
defnax: ok i had issue to create a useable retroshare service then not worked without thi s i will restore

src/jsonapi/jsonapi.cpp
```diff
@@ -283,6 +283,72 @@ JsonApiServer::JsonApiServer(): configMutex("JsonApiServer config"),
...
+			RsInit::LoadCertificateStatus retval;
+
+			if(!apiUser.empty() && badApiCredientalsFormat(apiUser, apiPass))
+			{
+				retval = RsInit::LoadCertificateStatus::ERR_UNKNOWN;
+			}
+			else
+			{
+				// If already logged in, treat as OK so we can register the token!
+				if(rsLoginHelper->isLoggedIn())
```
G10h4ck: Oh no! This is wrong and dangerous! With this any attacker can just use the `attemptLogin` method via API to inject auth token on already logged in profiles and use the API for whatever they want! Destroying all RS security with no effort at all!  Please read very carefully what AI generate, some things doesn't require C++ mastery to catch

G10h4ck (changes requested):

src/jsonapi/jsonapi.cpp
defnax: ok reverted it

PR #333, opened 2026-07-16
Added auth token fix for rs mobile

defnax:
# Walkthrough - Secure API Token Authentication

We have designed and implemented a safer authentication mechanism for RetroShare's JSON API. The PGP/profile password is no longer transmitted in the Authorization headers of REST requests, nor is it registered as the API token. Instead, we dynamically derive a secure API token client-side and authorize it upon login.

---

### 1. C++ Core (libretroshare)
- **[rsinit.h](file:///C:/Users/Ozan/Documents/GitHub/RetroShare/libretroshare/src/retroshare/rsinit.h)**: Marked `attemptLogin` as `manualwrapper` to exclude it from automatic API wrapper generation.
- **[jsonapi.cpp](file:///C:/Users/Ozan/Documents/GitHub/RetroShare/libretroshare/src/jsonapi/jsonapi.cpp)**: Implemented a manual wrapper for `/rsLoginHelper/attemptLogin` that deserializes optional `apiUser` and `apiPass` fields, calls `attemptLogin`, and registers the client's custom credentials using `authorizeUser(apiUser, apiPass)` if the login is successful.

### 2. Dart API Wrapper (Retroshare-Wrapper)
- **[retroshare.dart](file:///C:/Users/Ozan/Documents/GitHub/Retroshare-Wrapper/lib/src/retroshare.dart)**: Updated `requestLogIn` and `requestAccountCreation` to accept optional `apiUser` and `apiPass` arguments and include them in the request body parameters.

### 3. Mobile Client (rs-mobile)
- **[pubspec.yaml](file:///C:/Users/Ozan/Documents/GitHub/rs-mobile/pubspec.yaml)**:
  - Added the `crypto` package dependency.
- **[auth.dart](file:///auth.dart)**:
  - Added `deriveApiToken` helper function using `sha256(locationId:password)` to dynamically compute secure API credentials.
  - Modified `getinitializeAuth` to verify credentials using `deriveApiToken`.
  - Updated `login` and `signup` methods to derive the secure credentials and pass them during registration/login calls.

---

## Verification Results
- All components compiled successfully with zero errors.
- Dependencies were resolved correctly with `flutter pub get`.

### DIFF
--- a/src/jsonapi/jsonapi.cpp
+++ b/src/jsonapi/jsonapi.cpp
@@ -283,6 +283,65 @@ JsonApiServer::JsonApiServer(): configMutex("JsonApiServer config"),
 		} );
 	}, false);
 
+	registerHandler("/rsLoginHelper/attemptLogin",
+	                [this](const std::shared_ptr<rb::Session> session)
+	{
+		auto reqSize = session->get_request()->get_header("Content-Length", 0);
+		session->fetch( static_cast<size_t>(reqSize), [this](
+		                const std::shared_ptr<rb::Session> session,
+		                const rb::Bytes& body )
+		{
+			INITIALIZE_API_CALL_JSON_CONTEXT;
+
+			RsPeerId account;
+			std::string password;
+
+			// JSON API only
+			std::string apiUser;
+			std::string apiPass;
+
+			// deserialize input parameters from JSON
+			{
+				RsGenericSerializer::SerializeContext& ctx(cReq);
+				RsGenericSerializer::SerializeJob j(RsGenericSerializer::FROM_JSON);
+				RS_SERIAL_PROCESS(account);
+				RS_SERIAL_PROCESS(password);
+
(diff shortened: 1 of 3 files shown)


defnax:
this code is for https://github.com/RetroShare/rs-mobile/pull/38

G10h4ck (changes requested):
Looks good now, just a little nitpick, in the JSON API tag add `,unauthenticated` at the end, bieng it `manualwrapper` should not have any effect on the code, but makes it explicit for the reader that this method doesn't require API token authentication. After this lets hear @csoler opinion, better double check when touching this delicate things. Please also double check if it matches how `createLovationV2` deal with this, as this is basically porting the same mechanism we already have implemented for that method to this one too which makes sense.

G10h4ck (approved):
Seems good to me now, ready for @csoler pass ;-)

PR #334, opened 2026-07-18
Fix deprecated uppercase causes a crash on android

defnax:


csoler:
While I understand why it's always good to get rid of @jsonapi{development} on deprecated functions, I just do not understand why it causes a crash. Can you ellaborate?

defnax:
Ok i will send you latee the issue now on the way

defnax:
## 1. RESTbed "Resource would pollute namespace" Android Crash (Resolved)

### Root Cause
On Android startup, the RetroShare C++ daemon crashed inside `restbed::Service::publish` with the following exception:
```
E/libc++abi(22124): terminating due to uncaught exception of type std::invalid_argument: Resource would pollute namespace. Please ensure all published resources have unique paths.
```
In the RESTbed framework, publishing multiple resources that share the same path throws an exception.
In the C++ codebase, `libretroshare/src/retroshare/rsconfig.h` defines both the old deprecated uppercase methods and the new lowercase methods:
*   `GetMaxDataRates` and `getMaxDataRates`
*   `SetMaxDataRates` and `setMaxDataRates`

The Python script `jsonapi-generator.py` parses these and automatically generated wrappers for both, which mapped to:
*   `/rsConfig/GetMaxDataRates` and `/rsConfig/getMaxDataRates`
*   `/rsConfig/SetMaxDataRates` and `/rsConfig/setMaxDataRates`

Since RESTbed parses routes using regular expressions, these case-insensitive overlaps collided in the route table, throwing the namespace pollution error and crashing the daemon on Android.

defnax:
@G10h4ck  the changes are right?

G10h4ck (approved):

G10h4ck:
the changes are Ok, but I do think the error you are getting comes from another place, this methods have been there since forever it would be very strange they started failing now without being touched

defnax:
Yes the errors i get from mistake changes i think now is solved with #336

PR #336, opened 2026-07-19
Added auth token fix for retroshare mobile

defnax:
# Walkthrough - Secure API Token Authentication

We have designed and implemented a safer authentication mechanism for RetroShare's JSON API. The PGP/profile password is no longer transmitted in the Authorization headers of REST requests, nor is it registered as the API token. Instead, we dynamically derive a secure API token client-side and authorize it upon login.

---

### 1. C++ Core (libretroshare)
- **[rsinit.h](file:///C:/Users/Ozan/Documents/GitHub/RetroShare/libretroshare/src/retroshare/rsinit.h)**: Marked `attemptLogin` as `manualwrapper` to exclude it from automatic API wrapper generation.
- **[jsonapi.cpp](file:///C:/Users/Ozan/Documents/GitHub/RetroShare/libretroshare/src/jsonapi/jsonapi.cpp)**: Implemented a manual wrapper for `/rsLoginHelper/attemptLogin` that deserializes optional `apiUser` and `apiPass` fields, calls `attemptLogin`, and registers the client's custom credentials using `authorizeUser(apiUser, apiPass)` if the login is successful.

### 2. Dart API Wrapper (Retroshare-Wrapper)
- **[retroshare.dart](file:///C:/Users/Ozan/Documents/GitHub/Retroshare-Wrapper/lib/src/retroshare.dart)**: Updated `requestLogIn` and `requestAccountCreation` to accept optional `apiUser` and `apiPass` arguments and include them in the request body parameters.

### 3. Mobile Client (rs-mobile)
- **[pubspec.yaml](file:///C:/Users/Ozan/Documents/GitHub/rs-mobile/pubspec.yaml)**:
  - Added the `crypto` package dependency.
- **[auth.dart](file:///auth.dart)**:
  - Added `deriveApiToken` helper function using `sha256(locationId:password)` to dynamically compute secure API credentials.
  - Modified `getinitializeAuth` to verify credentials using `deriveApiToken`.
  - Updated `login` and `signup` methods to derive the secure credentials and pass them during registration/login calls.

---

## Verification Results
- All components compiled successfully with zero errors.
- Dependencies were resolved correctly with `flutter pub get`.

### DIFF
--- a/src/jsonapi/jsonapi.cpp
+++ b/src/jsonapi/jsonapi.cpp
@@ -283,6 +283,65 @@ JsonApiServer::JsonApiServer(): configMutex("JsonApiServer config"),
 		} );
 	}, false);
 
+	registerHandler("/rsLoginHelper/attemptLogin",
+	                [this](const std::shared_ptr<rb::Session> session)
+	{
+		auto reqSize = session->get_request()->get_header("Content-Length", 0);
+		session->fetch( static_cast<size_t>(reqSize), [this](
+		                const std::shared_ptr<rb::Session> session,
+		                const rb::Bytes& body )
+		{
+			INITIALIZE_API_CALL_JSON_CONTEXT;
+
+			RsPeerId account;
+			std::string password;
+
+			// JSON API only
+			std::string apiUser;
+			std::string apiPass;
+
+			// deserialize input parameters from JSON
+			{
+				RsGenericSerializer::SerializeContext& ctx(cReq);
+				RsGenericSerializer::SerializeJob j(RsGenericSerializer::FROM_JSON);
+				RS_SERIAL_PROCESS(account);
+				RS_SERIAL_PROCESS(password);
+
+				// JSON API only
+				RS_SERIAL_PROCESS(apiUser);
+				RS_SERIAL_PROCESS(apiPass);
+			}
+
+			RsInit::LoadCertificateStatus retval;
+
+			if(!apiUser.empty() && badApiCredientalsFormat(apiUser, apiPass))
+			{
+				retval = RsInit::LoadCertificateStatus::ERR_UNKNOWN;
+			}
+			else
+			{
+				// Call retroshare C++ API
+				retval = rsLoginHelper->attemptLogin(account, password);
+
+				// If login succeeded and custom API credentials were provided, authorize them!
+				if(retval == RsInit::OK && !apiUser.empty())
+				{
+					authorizeUser(apiUser, apiPass);
+				}
+			}
+
+			// serialize out parameters and return value to JSON
+			{
+				RsGenericSerializer::SerializeContext& ctx(cAns);
+				RsGenericSerializer::SerializeJob j(RsGenericSerializer::TO_JSON);
+				RS_SERIAL_PROCESS(retval);
+			}
+
+			// return them to the API caller
+			DEFAULT_API_CALL_JSON_RETURN(rb::OK);
+		} );
+	}, false);
+
 	registerHandler("/rsControl/rsGlobalShutDown",
 	                [](const std::shared_ptr<rb::Session> session)
 	{
--- a/src/jsonapi/jsonapi.cpp
+++ b/src/jsonapi/jsonapi.cpp
@@ -791,19 +791,33 @@ bool JsonApiServer::saveList(bool& cleanup, std::list<RsItem*>& saveItems)
 
 bool JsonApiServer::loadList(std::list<RsItem*>& loadList)
 {
+	/* This may run while the server is already listening and serving requests,
+	 * because in retroshare-service and Android the config manager only exists
+	 * after login, so protect the members shared with the request handlers. */
+	RS_STACK_MUTEX(configMutex);
+
 	for(RsItem* it : loadList)
     {
         JsonApiServerAuthTokenStorage *au=dynamic_cast<JsonApiServerAuthTokenStorage*>(it);
 
         if(au)
-            mAuthTokenStorage = *au;
+        {
+			/* Merge instead of replacing: tokens authorized before the config
+			 * gets loaded, like the web interface one supplied on the command
+			 * line, must not be wiped out. On conflict the already authorized
+			 * one wins, as it is the most recently supplied. */
+			for(auto& tk: au->mAuthorizedTokens)
+				mAuthTokenStorage.mAuthorizedTokens.insert(tk);
+        }
 
         JsonApiServerConfigItem *ac=dynamic_cast<JsonApiServerConfigItem*>(it);
 
         if(ac)
         {
-            mListeningPort = ac->mListeningPort;
-            mBindingAddress = ac->mBindingAddress;
+			/* Do not override what has been explicitly supplied on the command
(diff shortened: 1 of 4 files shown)


defnax:
@G10h4ck

defnax:
Hope merged soon i need continue on retroshare mobile

src/retroshare/rsinit.h
```diff
@@ -419,9 +419,18 @@ class RsLoginHelper
 
 	/**
 	 * @brief Normal way to attempt login
-	 * @jsonapi{development,unauthenticated}
+	 * @jsonapi{development,manualwrapper}
 	 * @param[in] account Id of the account to which attempt login
 	 * @param[in] password Password for the given account
+	 * unauthenticated
```
G10h4ck: move this line just after the one with the jsonapi tag

G10h4ck (changes requested):

src/retroshare/rsinit.h
defnax: I will do tonight after work

G10h4ck:
did you test it on at least a couple of platforms?

defnax:
i tesed with rs mobile flutter with windows desktop(w10) and android (Samsung) no issue

G10h4ck:
In that case seems good to me, @csoler can you take a look just in case?

csoler:
I will take a look asap. The sentence "..PGP password passed an an API token" is freaking scary.

src/retroshare/rsinit.h
```diff
@@ -419,9 +419,18 @@ class RsLoginHelper
 
 	/**
 	 * @brief Normal way to attempt login
-	 * @jsonapi{development,unauthenticated}
+	 * @jsonapi{development,manualwrapper}
+	 * unauthenticated
```
csoler, line 423: misplaced keyword?

defnax: @G10h4ck wanted this there not me

jolavillette:
Audit by Claude AI and apparently there is a real issue:

"Nice catch on the dead connectToConfigManager — confirmed on Linux (CMake/GCC, builds clean). With retroshare-service on a fresh profile, createLocationV2 with apiUser/apiPass then a clean shutdown: on master no jsonapi.cfg is written at all (the JsonApiServer is never registered with the config manager, so nothing is even attempted), and with your patch the token is saved and reloaded correctly.

One issue with the rsinit.cc part, which isn't in the walkthrough: it loads the config after startupWebServices() has applied the command line, and loadList() replaces mListeningPort, mBindingAddress and mAuthTokenStorage wholesale, so the persisted values now win over the CLI. Reproducer:

- run 1: retroshare-service -c /tmp/t -J 9095, create the location over the API, shut down cleanly — port 9095 goes into jsonapi.cfg
- run 2: retroshare-service -c /tmp/t -J 9096, log in over the API, then curl -u user:pass --data '{}' http://127.0.0.1:9096/rsJsonApi/listeningPort → returns 9095
- curl -u user:pass --data '{}' http://127.0.0.1:9096/rsJsonApi/restart → the API comes back on 9095 and 9096 is gone, i.e. the -J the operator asked for is silently dropped

In retroshare-gui the existing call at rsinit.cc:452 runs before those settings are applied, which is why the order is fine there. Re-applying the RsConfigOptions values after the load (or having the load not clobber explicitly supplied ones) would fix it. Two nits while there: the load runs twice per start — "Loaded auth tokens" is printed once by connectToConfigManager() and once by mConfigMgr->loadConfiguration() — and dynamic_cast<p3ConfigMgr*>(mConfigMgr) at line 1351 is a no-op, since mConfigMgr is already p3ConfigMgr* (p3face.h:186); the cast is only needed at line 452, where the source is the RsConfigMgr* base."

src/retroshare/rsinit.h
defnax: image

defnax: @G10h4ck

defnax:
@csoler ?

defnax:
@G10h4ck @csoler

defnax:
new review?

src/retroshare/rsinit.h
G10h4ck: good for me

G10h4ck (approved):
Seems good

defnax:
@csoler ?

src/jsonapi/jsonapi.cpp
```diff
@@ -283,6 +283,65 @@ JsonApiServer::JsonApiServer(): configMutex("JsonApiServer config"),
...
+				RS_SERIAL_PROCESS(password);
+
+				// JSON API only
+				RS_SERIAL_PROCESS(apiUser);
+				RS_SERIAL_PROCESS(apiPass);
+			}
+
+			RsInit::LoadCertificateStatus retval;
+
+			if(!apiUser.empty() && badApiCredientalsFormat(apiUser, apiPass))
```
csoler, line 317: I would use a || instead of a &&:

if(apiUser.empty() || badApiCredentialsFormat(...))

since apiUser is needed anyway to call autorizedUser() later on.

csoler (changes requested):

src/jsonapi/jsonapi.cpp
jolavillette: With || the login itself is refused when apiUser is empty, which makes the parameter mandatory on an existing endpoint. Two upstream repos already rely on it being optional:

- RetroShare/rs-mobile master calls RsLoginHelper.requestLogIn(currentAccount, password) (lib/provider/auth.dart:79) and treats anything but 0/1 as "WRONG PASSWORD", so users would get that with the correct password;

- RetroShare/Retroshare-Wrapper master already has the wrapper change merged, and it omits the fields from the JSON body when they are null (if (apiUser != null) 'apiUser': apiUser), so apiUser reaches the server empty on every plain login.

The && form keeps the previous behaviour when nothing is supplied — plain login, no token authorized — and only rejects a malformed pair. If the concern is apiPass being silently ignored when apiUser is missing, that case can be rejected on its own:

if( (apiUser.empty() && !apiPass.empty()) ||
    (!apiUser.empty() && badApiCredientalsFormat(apiUser, apiPass)) )
	retval = RsInit::LoadCertificateStatus::ERR_UNKNOWN;

Either way, ERR_UNKNOWN can't be told apart from "no such account" or "wrong password" by the caller, and LoadCertificateStatus has no code for a bad API token — so if apiUser is meant to become mandatory, the doc comment should say so explicitly.

G10h4ck: > I would use a || instead of a &&:
> 
> if(apiUser.empty() || badApiCredentialsFormat(...))
> 
> since apiUser is needed anyway to call autorizedUser() later on.

nope apiUser is not always needed in this case, it is passed only in case a new token is needed. So keep `! &&` form.

csoler: ok. @jolavillette for the record it looks like Claude is saying "the different implementations using this API work fine so we do not need to make the API more secure". This is a bit concerning because things should go the other way: do the client keep working if the API is made more restrictive.

PR #339, opened 2026-07-20
API for torstatus

defnax:


defnax:
@G10h4ck

G10h4ck:
Wrong, absolutely no need of manual wrappers for such simple methods

G10h4ck:
Also please do not paste all the verbose slop the AI spit. Write your genuine understanding of what is being done, that would be much more useful and a better use of out times ;-)

PR #344, opened 2026-07-24
Fix/reduce log noise

jolavillette:
Fix/reduce log noise

removes a lot of non relevant outputs from the console

### DIFF
--- a/src/chat/distributedchat.cc
+++ b/src/chat/distributedchat.cc
@@ -62,6 +62,40 @@ static const uint32_t 		MAX_MESSAGES_PER_SECONDS_PERIOD     =   10 ; // duration
 
 #define  EXTRACT_PRIVACY_FLAGS(flags) (ChatLobbyFlags(flags.toUInt32()) * (RS_CHAT_LOBBY_FLAGS_PUBLIC | RS_CHAT_LOBBY_FLAGS_PGP_SIGNED))
 
+// A single locally-banned identity can flood a lobby with thousands of items. We still drop every
+// one of them (that is the correct, expected behaviour, NOT an error), but one WARN per dropped item
+// buries the rest of the log. This rate-limits the logging: one line when the flood from a given
+// identity starts, then one summary line per identity per REPORT_INTERVAL carrying the number of
+// items dropped in between. No error is hidden: signature mismatches and other genuine problems are
+// logged separately and unconditionally.
+static void logBannedIdentityDrop(const RsGxsId& keyId)
+{
+    static const rstime_t REPORT_INTERVAL = 60; // seconds
+    static RsMutex banned_drop_log_mtx("bannedDropLog");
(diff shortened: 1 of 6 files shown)


src/pqi/pqissllistener.cc
```diff
@@ -866,26 +864,19 @@ int pqissllistener::finaliseConnection(int fd, SSL *ssl, const RsPeerId& peerId,
...
+		          + " from " + sockaddr_storage_tostring(remote_addr) + " => shutting down");
 		return -1;
 	}
 
-	out += "Found Matching Peer for Connection:";
-	out += sockaddr_storage_tostring(remote_addr);
-	out += "\npqissllistener => Passing to pqissl module!";
-	pqioutput(PQL_WARNING, pqissllistenzone, out);
-
-    std::cerr << "pqissllistenner::finaliseConnection() connected to " << sockaddr_storage_tostring(remote_addr) << std::endl;
```
csoler, line 888: could you keep this line?

src/gxs/rsgxsnetservice.cc
```diff
@@ -3749,7 +3749,10 @@ bool RsGxsNetService::encryptSingleNxsItem(RsNxsItem *item, const RsGxsCircleId&
 
 	if(!mCircles->recipients(destination_circle,destination_group,recipients))
 	{
-		std::cerr << "  (EE) Cannot encrypt transaction: recipients list not available. Should re-try later." << std::endl;
+		// Not an error: the circle membership may simply not be cached yet. The caller retries later.
```
csoler: This one can be seen as an error. Generally speaking if the circle known, all the identities should be there as well.

src/chat/distributedchat.cc
```diff
@@ -62,6 +62,40 @@ static const uint32_t 		MAX_MESSAGES_PER_SECONDS_PERIOD     =   10 ; // duration
...
+// identity starts, then one summary line per identity per REPORT_INTERVAL carrying the number of
+// items dropped in between. No error is hidden: signature mismatches and other genuine problems are
+// logged separately and unconditionally.
+static void logBannedIdentityDrop(const RsGxsId& keyId)
+{
+    static const rstime_t REPORT_INTERVAL = 60; // seconds
+    static RsMutex banned_drop_log_mtx("bannedDropLog");
+    static std::map<RsGxsId,std::pair<uint32_t,rstime_t> > stats; // id -> (drops since last report, last report time)
+
+    RS_STACK_MUTEX(banned_drop_log_mtx);
```
csoler: Make sure we actually need a mutex here. All the calls come from the loop that receives chat items.

jolavillette:
done — connection line restored; the missing-recipients case is logged unconditionally again (one-line RsWarn with the circle id); mutex removed, both callers are indeed in the single-threaded receive path.

PR #346, opened 2026-07-24
API for torStatus

defnax:


defnax:
code for statusbar on webui
https://github.com/RetroShare/RSNewWebUI/pull/120/changes/e61837d550e23c4d954e0d7955ad47c0fcdf3541

      // Check if node uses automated Tor management via /rsAccounts/isTorAuto (same as Qt)
      rs.rsJsonApiRequest('/rsAccounts/isTorAuto', {}).then((autoRes) => {
        const isAuto = autoRes && (autoRes.retval || (autoRes.body && autoRes.body.retval));
        if (isAuto) {
          Promise.all([
            rs.rsJsonApiRequest('/rsTor/torStatus', {}),
            rs.rsJsonApiRequest('/rsTor/torConnectivityStatus', {}),
          ]).then(([torRes, connRes]) => {
            const torStatus        = torRes  && torRes.body  ? torRes.body.retval  : (torRes  && torRes.retval !== undefined ? torRes.retval  : 0);
            const connStatus       = connRes && connRes.body ? connRes.body.retval : (connRes && connRes.retval !== undefined ? connRes.retval : 0);
            const torControlOk     = connStatus === 6; // HIDDEN_SERVICE_READY
            const torReady         = torStatus === 2;   // READY

            if (torReady && torControlOk) {
              State.torProxyOk     = true;
              State.torChecking    = false;
            } else if (torStatus === 1 || connStatus === 0 || connStatus === 1) {
              // OFFLINE, ERROR, or NOT_CONNECTED
              State.torProxyOk     = false;
              State.torChecking    = false;
            } else if (connStatus >= 2 && connStatus <= 5) {
              // CONNECTING, SOCKET_CONNECTED, AUTHENTICATING, AUTHENTICATED
              State.torProxyOk     = null;
              State.torChecking    = true;
            } else {
              // UNKNOWN / default
              State.torProxyOk     = null;
              State.torChecking    = false;
            }
            m.redraw();
          }).catch(() => {
            State.torProxyOk     = true;
            State.torChecking    = false;
            m.redraw();
          });
        } else {
          // Manual Tor / I2P proxy node
          State.torProxyOk     = true;
          State.torChecking    = false;
          m.redraw();
        }

src/tor/TorManager.cpp
```diff
@@ -846,36 +848,43 @@ uint16_t RsTor::socksPort()
...
+    case TorControl::Connecting:         return RsTorConnectivityStatus::CONNECTING;
+    case TorControl::SocketConnected:    return RsTorConnectivityStatus::SOCKET_CONNECTED;
+    case TorControl::Authenticating:     return RsTorConnectivityStatus::AUTHENTICATING;
+    case TorControl::Authenticated:      return RsTorConnectivityStatus::AUTHENTICATED;
+    case TorControl::HiddenServiceReady: return RsTorConnectivityStatus::HIDDEN_SERVICE_READY;
+    case TorControl::Unknown:            return RsTorConnectivityStatus::UNKNOWN;
     }
 }
 
+RsTorConnectivityStatus RsTor::torConnectivityStatus()
```
csoler, line 883: Why did you need to split this function into two functions ??

csoler (commented):
look ok and works. I just don't understand the need for splitting the status functions into 2 functions.

src/tor/TorManager.cpp
defnax: Functionally yes, but there is a small code duplication: the same static torConnectivityStatus(Tor::TorControl::Status t) body is copy-pasted identically in both TorControl.cpp and TorManager.cpp. A cleaner alternative would be to put the helper in a shared internal header (e.g. tor/TorStatusHelper.h) and include it in both .cpp files — that way the logic lives in one place.

Summary: The split was necessary because TorControl.cpp needs the same enum-to-RsTorConnectivityStatus conversion logic but cannot call the RsTor:: member function (it would create a circular dependency with TorManager). The static helper is duplicated across both translation units to solve this. A minor improvement would be to deduplicate it into a shared private header.

defnax: ### Why it was needed — the key reason

Look at [TorControl.cpp L64-L78](file:///c:/Users/ozzyd/Documents/GitHub/libretroshare/src/tor/TorControl.cpp#L64-L88): **the exact same static helper now also exists in `TorControl.cpp`**, where it is used in `TorControl::setStatus()` (line 107) and other places to fill events.

`TorControl.cpp` **cannot call `RsTor::torConnectivityStatus()`** because that would:
- Call `RsTor::instance()` → which creates a `TorManager`
- `TorControl` is a member of `TorManager` — a circular dependency
- `TorControl.cpp` sits at a lower layer and must not depend on `TorManager`/`RsTor`

So the conversion logic (`switch` over the enum) **needs to be available in both compilation units** (`TorControl.cpp` and `TorManager.cpp`) without either depending on the other. The split makes the public API (`RsTor::torConnectivityStatus()`) a thin wrapper, while the reusable static helper can live independently in each `.cpp` file.

PR #341, opened 2026-07-22
API for get Avatar V2

defnax:

PR #350, opened 2026-07-27
Add rsChats/getAvatar and rsChats/getOwnAvatar to the JSON API

jolavillette:
Add rsChats/getAvatar and rsChats/getOwnAvatar to the JSON API.

Supersedes #341 and #321, which encoded base64 by hand inside
p3ChatService.

Both methods return the avatar as an RsGxsImage, the container already
used to carry identity avatars over the JSON API. Its serial_process()
goes through RsTypeSerializer::RawMemoryWrapper, so the base64 wrapping
of the binary payload is produced by the serialization layer itself
rather than by the service, as requested in #1977. The resulting
representation is the one API clients already handle for identity
avatars:

    $ curl -u $API_TOKEN -d '{"pid":"<ssl id>"}' \
        $API_BASE_URL/rsChats/getAvatar
    {"avatar":{"mData":{"base64":"/9j/4AAQSkZJRgABAQ..."}},"retval":true}

When the avatar of a peer is not known yet, getAvatar() returns false
and asks the peer for it, so a later call may succeed.

The existing getAvatarData() is kept as is for its existing callers: it
hands out a raw malloc'ed buffer and an int size, which is neither
serializable nor usable from the JSON API, and the GUI still relies on
it. Its "create bookkeeping entry then ask the peer, at most once a
minute" part is factored out into locked_requestAvatar() and shared
with the new getAvatar(), and its out parameters are now explicitly
cleared when no avatar is available.

### DIFF
--- a/src/chat/p3chatservice.cc
+++ b/src/chat/p3chatservice.cc
@@ -1501,6 +1501,27 @@ std::string p3ChatService::getCustomStateString(const RsPeerId& peer_id)
 	return std::string() ;
 }
 
+void p3ChatService::locked_requestAvatar(const RsPeerId& peer_id)
+{
+	/* Create placeholder to store request time */
+	std::map<RsPeerId,AvatarInfo *>::const_iterator it = _avatars.find(peer_id) ;
+
+	if(it == _avatars.end())
+	{
+		_avatars[peer_id] = new AvatarInfo();
+		it = _avatars.find(peer_id);
+	}
+
+	time_t now = time(NULL);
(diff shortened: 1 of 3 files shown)


defnax:
We need it for webui & on mobile app this api

defnax:
@csoler @G10h4ck

defnax:
@csoler @G10h4ck

defnax:
this is now forgettten @G10h4ck we has now a nice new webui but not jet Avatar retrieve

PR #351, opened 2026-07-27
GXS: fix four causes of slow channel message loading

jolavillette:
GXS: fix four causes of slow channel message loading

works with retroshare-gui pr/3280
https://github.com/RetroShare/RetroShare/pull/3280

Opening a channel with 6409 messages took several seconds. It now takes about half a second on the same profile. Four independent defects, each measured before and after.

One SQL query per message. A request for a whole group carries an empty message id set, meaning "everything". RsGxsDataAccess::getMsgData() expanded it into the explicit list of all 6409 ids before reaching the store, which forced the per-message query path: 6409 sqlite3_prepare_v2 cycles against an encrypted database. The single-query path existed and was never taken.

Superseded post versions were read in full. A channel keeps every version of every edited post; only the latest is displayed. sortPosts() read them all and discarded the old ones after decrypting and deserialising them — 195 MB read to display 23 MB worth of posts. Version chains are now resolved on the metas, which are small and cached, and only the retained versions are read.

No move semantics. RsGxsImage, RsMsgMetaData, RsGxsGenericMsgData and RsGxsChannelPost all declare destructors, which suppresses the implicit move operations, so every std::move() silently became a malloc+memcpy of the thumbnail. The post array was deep copied at four hand-off points.

A query planner trap. With no ANALYZE data sqlite estimates an equality on a non-unique index at ~10 rows, so grpId=? AND msgId IN (...) picked the group index and rescanned the whole group once per batch. Selecting on the message id alone — it is the primary key, unique table wide — fixes it.

No schema change, no public API change, no behaviour change: returned posts, comments, votes and comment counts are identical on every path.

The first commit adds opt-in profiling (RS_GXS_PROFILE=0), which is what makes all of the above verifiable. Happy to drop it if you would rather not carry it.

Review focus: applyPostVersions() reproduces two behaviours of sortPosts() — comments hanging off a superseded version must still be counted on the retained post, and the mOrigMsgId normalisation that updateSinglePost() matches on must be carried over.

Companion PR in the GUI repository: pr/3280. It needs this one merged first.

### DIFF
--- a/src/services/p3gxschannels.cc
+++ b/src/services/p3gxschannels.cc
@@ -1552,8 +1552,14 @@ bool p3GxsChannels::getChannelAllContent( const RsGxsGroupId& channelId,
     std::vector<RsMsgMetaData> post_metas;
     std::set<RsGxsMessageId> wanted_msgs;
 
+    // Only mThreadId tells posts apart from the rest: a post never has one
+    // (createChannelPost leaves it null), while a comment always carries the id
+    // of the post it belongs to and a vote the id of the post being commented.
+    // mParentId is not usable here: comments written under the old paradigm have
+    // a null one.
+
     for(auto& m: metas)
-        if(m.mThreadId.isNull() && m.mParentId.isNull())
+        if(m.mThreadId.isNull())
             post_metas.push_back(m);
         else
             wanted_msgs.insert(m.mMsgId);	// comments and votes are all kept
--- a/src/services/p3gxschannels.cc
+++ b/src/services/p3gxschannels.cc
@@ -609,14 +609,12 @@ bool p3GxsChannels::groupShareKeys(
  * at the moment - fix it up later
  */
 
-bool p3GxsChannels::getPostData( const uint32_t& token, std::vector<RsGxsChannelPost>& msgs,
-                                 std::vector<RsGxsComment>& cmts,
-                                 std::vector<RsGxsVote>& vots)
+bool p3GxsChannels::convertMsgItems( const uint32_t& token,
+                                     std::vector<RsGxsChannelPost>& msgs,
+                                     std::vector<RsGxsComment>& cmts,
+                                     std::vector<RsGxsVote>& vots,
+                                     long& getmsgdata_ms, long& convert_ms )
 {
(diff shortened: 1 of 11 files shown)


csoler (changes requested):
Could you plz remove all the extra profiling code? Of course you may keep the profiling classes themselves.  This will make the PR much easier to read.

jolavillette:
done

src/services/p3gxschannels.cc
```diff
@@ -1511,14 +1534,148 @@ bool p3GxsChannels::getChannelAllContent( const RsGxsGroupId& channelId,
...
+    std::vector<RsMsgMetaData> metas;
+
+    if(!getContentSummaries(channelId,metas))
+        return false;
+
+    std::vector<RsMsgMetaData> post_metas;
+    std::set<RsGxsMessageId> wanted_msgs;
+
+    for(auto& m: metas)
+        if(m.mThreadId.isNull() && m.mParentId.isNull())
```
csoler: For a top level comment:
- mThreadId is equal to the actual post the comment refers to, so it's never null, 
- mParentId can be actually null if following the old comment paradigm (in the code now, it's never null)
For votes, 
- mThreadId is the post id which comment is voted,
- mParentId is the comment that is voted. 

For posts,
- mParentId is always null (see p3gxschannels.cc:2021)
- mThreadId is always null (see p3gxschannels.cc:2015)

So the test above should be safer to only look for mThreadId.isNull() in order to accurately discard what is not a post.

csoler (changes requested):

src/services/p3gxschannels.cc
jolavillette: Right, thanks — fixed in 2a2a1e241.

mParentId proves nothing here since old-paradigm comments have a null one, so the test is now on mThreadId alone, with a comment recording the taxonomy you described.

For the record, the composite test comes from getChannelStatistics() (p3gxschannels.cc:1721), which I used as a model, and the same one is in p3wire.cc:1837 for pulses. I left both untouched as they are out of this PR's scope — happy to align them here or in a separate PR, your call.

PR #352, opened 2026-07-27
Fix tor

thunder2:


csoler:
Is there a reason why you would need this? (moving the tor data directory into each separate instance data dir). If so, then it needs to be done consistently (in main.cpp of retroshare-gui). Also  in some places, the tor directory is set to the base retroshare data dir because there is no location involved (e.g. in retroshare friend server main function).

thunder2:
Currently we cannot run multiple tor instances simultaneously. The changes in rsinit.cc is for the Retroshare service. The changes for the gui is in PR 3281. I have not checked the friend server.

csoler:
I suppose that the friend server should also have its own tor directory where it places its own data.

thunder2:
I found it. The friend server uses the same directory for the tor data as the one the hidden_service is stored. This should correspond to the account directory.

PR #355, opened 2026-07-29
GXS: warm up the message meta caches with one scan instead of one query per group

jolavillette:
Builds on #351 (same function, and reuses the profiler introduced there). One commit on top of it.

## Problem

Reading the meta of a whole group runs `SELECT ... WHERE grpId=?`, which `INDEX_MESSAGES_GRPID` serves with one row lookup per message. The payload blob lives in the same row, so those lookups are scattered over the whole file. Warming up the cache of N groups therefore costs N passes of random I/O over a database that is hundreds of megabytes, and the cost grows linearly with the number of groups.

On a node subscribed to hundreds of forums this shows up as tens of seconds before the unread counters settle at startup, with the biggest groups taking seconds each on their own (26 s measured for a 9573-message forum on a spinning disk).

## Fix

When more than one group still needs a cold full read, read the meta of every message in a **single sequential scan** and fill every per-group cache from it. The file is then read in physical order, and warming up all the groups costs one pass whatever their number.

Measured on a synthetic database of the same shape and size as a real `gxsforums_db` (235 MB, 23 KB rows, 20 groups, cold page cache enforced between runs):

```
20 per-group queries   29449 ms
one sequential scan     2146 ms      13.7x
```

The per-group path grows with the number of groups; the scan does not.

## Scope

- No schema change, no format change, no migration, no public API change. Same columns, same cache contents, same values returned to callers.
- The scan reports itself through the opt-in profiler (`RS_GXS_PROFILE`), so the gain is checkable on a real profile:
  `GXS-PROF loadAllMsgMetaInOneScan db=gxsforums_db groups=571 metas=48213 in 2100ms`

## Trade-off worth reviewing

The scan fills the cache for groups that were not requested yet. That is the same memory the cache reaches as soon as those groups are browsed, but it is reached up front instead of progressively. The trigger is deliberately conservative (more than one group needing a cold full read, and only once per database), but if you would rather bound it explicitly — by message count, or by only caching the requested groups — say so and I will adjust.

### DIFF
--- a/src/gxs/rsdataservice.cc
+++ b/src/gxs/rsdataservice.cc
@@ -131,6 +131,8 @@ RsDataService::RsDataService(const std::string &serviceDir, const std::string &d
 
     mDb = new RetroDb(mDbPath, RetroDb::OPEN_READWRITE_CREATE, key);
     mUseCache = true;
+    mMsgMetaDataCache_ContainsAllDatabase = false;
+    mMsgMetaDataCache_ColdFullReads = 0;
 
     initialise(isNewDatabase);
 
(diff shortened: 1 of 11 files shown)


jolavillette:
Closing: measured on a real profile, this makes things **worse**, and my benchmark was wrong.

On a real `gxsforums_db` (224 MB) the single scan took **78 s**, where the previous per-group reads totalled 51-65 s. Same code and same disk, the three databases behave completely differently:

```
gxschannels_db   18630 metas    8482 ms   608 MB  ->  72 MB/s
posted_db         5991 metas   12454 ms   305 MB  ->  24 MB/s
gxsforums_db     23472 metas   78065 ms   224 MB  ->  2.9 MB/s
```

A full table scan is only sequential if the rows are laid out in physical order. A forum database is constantly purged (storage period), so freed pages get reused out of order and the large payload blobs live in overflow chains scattered across the file. The scan then degenerates into random I/O, which is what 2.9 MB/s means.

My benchmark showed 13.7x because the synthetic database had just been written in one go and was perfectly contiguous -- it did not represent a long-lived database. That is a flaw in the measurement, not a detail.

Leaving this closed rather than tuning the trigger: the premise itself does not hold.

PR #345, opened 2026-07-24
Fix/remote dir cleanup delay

jolavillette:
Fix/remote dir cleanup delay

p3FileDatabase::cleanup() never purges a friend's cached file list, even after months offline. Two fixes:

Delays off by 24× (file_sharing_defaults.h): 86400 is already a day, but the thresholds had an extra *24 → the "60 / 5 days" comments were really 1440 / 120 days. Dropped the *24.

"Empty" now matches the GUI: cleanup used getDirectoryRecursModTime()==0, so a folder-only tree (0 files) stayed on the 60-day timer while Friends Files shows it Empty. Now uses dir_cumulated_files==0 (new getDirectoryCumulatedFileCount()) → pruned on the 5-day timer.

Timer is keyed on lastcontact (real connect/disconnect, persisted), so reconnecting peers are unaffected.

PR #356, opened 2026-08-03
Fix build break from #345: add missing DirEntry::dir_cumulated_files

jolavillette:
#345 used `DirEntry::dir_cumulated_files` in the new `DirectoryStorage::getDirectoryCumulatedFileCount()`, but the commit introducing that member was on a separate local branch and was not included in the PR — so master no longer compiles (first caught by the Android GitLab CI, but it breaks every platform):

```
src/file_sharing/directory_storage.cc:147:21: error: no member named 'dir_cumulated_files' in 'InternalFileHierarchyStorage::DirEntry'
```

This adds the missing piece: `dir_cumulated_files` is maintained alongside `dir_cumulated_size` in `recursUpdateCumulatedSize()`, which already runs on hierarchy load and on `checkSave()` for both local and remote directories, so the count used by the #345 cleanup logic is always up to date. Also exposed as `DirDetails::count` (cumulative file count of the hierarchy below), which the GUI can use later to display file counts.

Sorry for the breakage — the PR was developed and tested on a tree that already carried this commit.

### DIFF
--- a/src/file_sharing/dir_hierarchy.cc
+++ b/src/file_sharing/dir_hierarchy.cc
@@ -674,15 +674,21 @@ uint64_t InternalFileHierarchyStorage::recursUpdateCumulatedSize(const Directory
     DirEntry& d(*static_cast<DirEntry*>(mNodes[dir_index])) ;
 
     uint64_t local_cumulative_size = 0;
+    uint32_t local_cumulative_files = 0;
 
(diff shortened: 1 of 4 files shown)


defnax:
yes i need now for rs mobile this fix
https://gitlab.com/RetroShare/libretroshare/-/jobs/15684017419

csoler:
Thx. I was wrong assuming that 345 was compiling fine.

PR #360, opened 2026-08-06
fix(rsaccounts): locate the data directory relative to the executable on Linux

jolavillette:
`RS_DATA_DIR` is baked in at compile time, so a relocatable tree (AppImage, portable directory, staged install) can never find its own data: the binary reads the host's system-wide path instead, or nothing at all.

On Linux (excluding Android) prefer `<exedir>/../share/retroshare` when that directory exists, and fall back to `RS_DATA_DIR` otherwise. System-wide installs and build trees are unaffected.

This covers every file served from the data directory; the web UI in particular, which `p3webui` resolves as `PathDataDirectory() + "/webui/"`.

Verified on Linux: a binary placed in a relocated tree resolves its own data directory, while the same binary in a build tree still reports `RS_DATA_DIR`. Full Qt5 build green.

### DIFF
--- a/src/rsserver/rsaccounts.cc
+++ b/src/rsserver/rsaccounts.cc
@@ -794,6 +794,35 @@ static bool checkAccount(const std::string &accountdir, AccountDetails &account,
 	//#include <CFBundle.h>
 #endif
 
+#if defined(__linux__) && !defined(__ANDROID__)
+#include <limits.h>
+#include <stdlib.h>
+#include <unistd.h>
+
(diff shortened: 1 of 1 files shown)


csoler:
The way this PR handles things is not appropriate imo because looking for a / in the exe path looks rather fragile. 

I think the solution is way simpler: in appimage for instance, just make the launcher use "retroshare -c ./data/" or anything that suits the wanted configuration.

PR #365, opened 2026-08-11
Added API for getBoardPostSummaries

defnax:
is required to load boards post on webui faster
webui on channels is faster, it using the getContentSummaries

chatgpt says (5.6)
> Without the libretroshare addition, WebUI only has:
> 
> - getBoardAllContent: returns every post, comment, vote, and image in one large response.
> - getBoardContent: returns selected items, but requires message IDs that WebUI does not yet know.
> 
> So WebUI cannot request only the newest 25 posts without first receiving their IDs. It can improve the experience with loading indicators, caching, or prefetching, but the initial large download would still take 20–30 seconds.
> The small getBoardPostSummaries() core API is necessary to make the first page genuinely load faster.

### DIFF
--- a/src/retroshare/rsposted.h
+++ b/src/retroshare/rsposted.h
@@ -233,6 +233,17 @@ public:
 	        std::vector<RsGxsVote>& votes ) = 0;
 
 	/**
+	 * @brief Get lightweight summaries for top-level posts in a board.
+	 * @jsonapi{development}
+	 * @param[in] boardId id of the board of which post summaries are requested
+	 * @param[out] summaries storage for post metadata, excluding comments and votes
+	 * @return false if something failed, true otherwise
+	 */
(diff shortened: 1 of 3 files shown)


defnax:
https://github.com/RetroShare/RSNewWebUI/pull/121/changes/88aaf7cf927e025442390ffb9b37286473df448e

Implemented true progressive loading for large Boards
Changes:
* Added getBoardPostSummaries() to libretroshare.
* Returns only lightweight top-level post metadata, excluding comments and votes.
* WebUI sorts summaries newest-first.
* Complete posts download in batches of 25.
* First page renders before older posts finish loading.
* Older RetroShare cores retain the existing getBoardAllContent fallback.

jolavillette:
please review pr/365, needed by RSNewWebUI pr/121

PR #368, opened 2026-08-15
CMake CI: build libretroshare on Linux, Windows and macOS

jolavillette:
libretroshare has no CI that compiles it. This adds three CMake jobs — Ubuntu, Windows UCRT64, macOS — all green:

https://github.com/jolavillette/libretroshare/actions?query=branch%3Aci%2Fcmake

No submodules, no super-project: the top level `CMakeLists.txt` pulls rnp, OpenPGP-SDK, BitDHT, rapidjson, restbed and udp-discovery-cpp through FetchContent, so a checkout plus the system libraries is enough. Options follow the super-project's Linux CMake job, minus what does not exist here (GUI, service, friendserver, plugins).

Nothing outside `.github/workflows/` is touched.

### Three things the standalone build hits, that the super-project never does

It always takes the `../supportlibs/...` branch of the dependency lookups, so the FetchContent branch had never run anywhere. Worked around in the workflows, but they are worth fixing in `CMakeLists.txt`:

1. **rapidjson**: `FetchContent_MakeAvailable(rapidjson)` adds no include directory, so `rapidjson/document.h` is not found. Workaround: install the system package.
2. **restbed**: the patch step runs `sed -i -e ...`, which is GNU syntax; BSD sed reads `-e` as a backup suffix. Workaround: GNU sed first in PATH on macOS.
3. **udp-discovery-cpp**: fetched at `GIT_TAG "origin/master"` — a moving third-party tip. It sets `CXX_STANDARD 98` on its target while its sources need C++11; UCRT64 refuses. Workaround: `RS_BRODCAST_DISCOVERY=OFF` on Windows only, noted in the file. Pinning a tag would fix it properly.

### DIFF
--- a/.github/workflows/macos-cmake.yml
+++ b/.github/workflows/macos-cmake.yml
@@ -50,7 +50,7 @@ jobs:
         brew install
         cmake ninja pkg-config
         openssl@3 botan@2 sqlcipher miniupnpc json-c
-        bzip2 zlib doxygen gnu-sed
+        bzip2 zlib doxygen gnu-sed rapidjson
 
     # The restbed patch step of CMakeLists.txt runs `sed -i -e ...`, which is
     # GNU syntax: BSD sed reads the argument after -i as a backup suffix and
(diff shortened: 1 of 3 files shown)


defnax:
by the way this one is not functional needs to be removed `ci-mingw64.yml`

csoler:
I'm surprised that we need a flag=ON for rnp. It should be the default now.

PR #370, opened 2026-08-16
build: wipe the doxygen xml dir before regenerating the jsonapi bindings

jolavillette:
A build directory that predates the removal or rename of a public header keeps failing with:

```
jsonapi-includes.inl:11:10: fatal error: retroshare/rsmsgs.h: No such file or directory
```

`rsmsgs.h` was removed by fd79225f8 ("started separating rsMsgs into rsMail and rsChats"), and grepping the sources for it finds nothing — because it is not in the sources. Doxygen never purges its `OUTPUT_DIRECTORY`, so `rsmsgs_8h.xml` is still sitting in the build tree, and both generators emit one `#include` per `*_8h.xml` file they find there, built from that XML's `<location file=...>`. The dead header is re-included on every rebuild, and `jsonapi-wrappers.inl` gets the matching dead wrappers.

`make clean` does not cure it: `genjsonapi.clean` lists only the two `.inl` files, never the `xml/` directory. The clean deletes the outputs, doxygen reruns, and the same `#include` comes back from the surviving XML. The only workaround is deleting the whole build directory, which is not something the error message suggests.

So delete the `*_8h.xml` files right before invoking doxygen, in both build systems. Only those files drive the generators (`jsonapi-generator.cpp` iterates `*8h.xml`, `jsonapi-generator.py` filters on `endswith("8h.xml")`); every other XML they open is reached through a live header's XML, which doxygen has just regenerated. This costs nothing: doxygen re-parses its whole input on every run anyway, it just stops reusing the stale header files.

`cmake -E rm` does not expand wildcards, so the CMake side does the glob in a small script generated next to the doxyfile.

Reproduced by dropping a single crafted `rsmsgs_8h.xml` into an otherwise up-to-date xml directory — the generator emits `#include "retroshare/rsmsgs.h"` again; with the deletion in place the output is clean.

Reported by Cyril on the dev chat.

### DIFF
--- a/src/libretroshare.pro
+++ b/src/libretroshare.pro
@@ -1029,9 +1029,14 @@ rs_jsonapi {
     genjsonapi.clean = $${WRAPPERS_INCL_FILE} $${WRAPPERS_REG_FILE}
     genjsonapi.CONFIG += target_predeps combine no_link
     genjsonapi.variable_out = HEADERS
+    # Doxygen never purges its output directory, and the generator emits one
+    # #include per *_8h.xml file it finds there. Without removing them first, the
+    # XML of a public header that has been removed or renamed survives forever and
+    # keeps being turned into an #include of a file that no longer exists.
     win32-g++:isEmpty(QMAKE_SH) {
         genjsonapi.commands = \
             $(CHK_DIR_EXISTS) $$shell_path($$JSONAPI_GENERATOR_OUT) $(MKDIR) $$shell_path($${JSONAPI_GENERATOR_OUT}) $$escape_expand(\\n\\t) \
+            -$(DEL_FILE) $$shell_path($${JSONAPI_GENERATOR_OUT}/xml/*_8h.xml) $$escape_expand(\\n\\t) \
             $(COPY_FILE) $$shell_path($${DOXIGEN_CONFIG_SRC}) $$shell_path($${DOXIGEN_CONFIG_OUT}) $$escape_expand(\\n\\t) \
             echo OUTPUT_DIRECTORY=$${JSONAPI_GENERATOR_OUT} >> $$shell_path($${DOXIGEN_CONFIG_OUT}) $$escape_expand(\\n\\t) \
             echo INPUT=$${DOXIGEN_INPUT_DIRECTORY} >> $$shell_path($${DOXIGEN_CONFIG_OUT}) $$escape_expand(\\n\\t) \
@@ -1039,6 +1044,7 @@ rs_jsonapi {
     } else {
         genjsonapi.commands = \
             mkdir -p $${JSONAPI_GENERATOR_OUT} && \
+            rm -f $${JSONAPI_GENERATOR_OUT}/xml/*_8h.xml && \
             cp $${DOXIGEN_CONFIG_SRC} $${DOXIGEN_CONFIG_OUT} && \
             echo OUTPUT_DIRECTORY=$${JSONAPI_GENERATOR_OUT} >> $${DOXIGEN_CONFIG_OUT} && \
             echo INPUT=$${DOXIGEN_INPUT_DIRECTORY} >> $${DOXIGEN_CONFIG_OUT} && \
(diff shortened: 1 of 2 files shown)


src/libretroshare.pro
```diff
@@ -1029,16 +1029,22 @@ rs_jsonapi {
...
             $(CHK_DIR_EXISTS) $$shell_path($$JSONAPI_GENERATOR_OUT) $(MKDIR) $$shell_path($${JSONAPI_GENERATOR_OUT}) $$escape_expand(\\n\\t) \
+            -$$QMAKE_DEL_TREE $$shell_path($${JSONAPI_GENERATOR_OUT}/xml) $$escape_expand(\\n\\t) \
             $(COPY_FILE) $$shell_path($${DOXIGEN_CONFIG_SRC}) $$shell_path($${DOXIGEN_CONFIG_OUT}) $$escape_expand(\\n\\t) \
             echo OUTPUT_DIRECTORY=$${JSONAPI_GENERATOR_OUT} >> $$shell_path($${DOXIGEN_CONFIG_OUT}) $$escape_expand(\\n\\t) \
             echo INPUT=$${DOXIGEN_INPUT_DIRECTORY} >> $$shell_path($${DOXIGEN_CONFIG_OUT}) $$escape_expand(\\n\\t) \
             doxygen $$shell_path($${DOXIGEN_CONFIG_OUT}) $$escape_expand(\\n\\t)
     } else {
         genjsonapi.commands = \
             mkdir -p $${JSONAPI_GENERATOR_OUT} && \
+            rm -rf $${JSONAPI_GENERATOR_OUT}/xml && \
```
csoler: Could you rather use something like "rm $${JSONAPI_GENERATOR_OUT}/xml/*_8h.xml" ? That line looks dangerous, if something ever happens to the JSONAPI_GENERATOR_OUT variable

csoler (changes requested):

src/libretroshare.pro
jolavillette: Agreed, done — no more directory removal anywhere.

Both generators only ever pick up the header XML (`jsonapi-generator.cpp` iterates `*8h.xml`, `jsonapi-generator.py` filters on `endswith("8h.xml")`), so deleting just those is enough: every other XML they open is reached through a live header'''s XML, which doxygen has just regenerated.

- sh branch: `rm -f $${JSONAPI_GENERATOR_OUT}/xml/*_8h.xml` (`-f` so an unexpanded glob is not an error)
- win32 no-sh branch: `-$(DEL_FILE) ... /xml/*_8h.xml`
- CMake: `cmake -E rm` does not expand wildcards, so the glob is done by a tiny script generated next to the doxyfile (`file(GLOB ...)` + `file(REMOVE ...)`, guarded against an empty match).

Force-pushed.

csoler:
Given this change, does the compilation always regenerate the xml files, or does it keep the existing ones when the .h files haven't changed?

csoler:
Maybe that cleaning step should be in the make clean, not in the compilation.

jolavillette:
Doxygen has no incremental mode and never purges: every run re-parses the whole INPUT and rewrites the XML of everything it parses, leaving behind the XML of whatever is gone (checked with 1.9.8). That leftover is the bug.

The deletion is a command inside the same rule as doxygen, so it runs if and only if doxygen runs, and it only removes files doxygen is about to rewrite anyway, plus those leftovers. No extra work, no extra run.

On `make clean`: it cannot be cured there, and it is not cured there today either. Reproduced on a two-header sample with the real `jsonapi-generator.py` and `jsonapi-generator-doxygen.conf`:

1. two headers -> both `#include`s generated
2. delete `rsbbb.h`, delete both `.inl` (which is all `genjsonapi.clean` does), rerun -> `#include "retroshare/rsbbb.h"` is back, plus the wrappers for its methods
3. same state, but purging the `*_8h.xml` first -> only the live include, no dead wrapper

Step 2 is a clean build tree as far as the clean rule is concerned; someone still has to know to delete the xml directory by hand. The stale XML is an *input* of the generation step, so purging it belongs to that step.

If you want `make clean` to drop the xml as well, I'll add that on top.

PR #374, opened 2026-08-17
Android: pin the rnp crypto backend to openssl for libretroshare too

jolavillette:
`build_librnp` builds and installs rnp into the Android sysroot with `-DCRYPTO_BACKEND=openssl`, but libretroshare's own configure can build rnp a **second** time, inline: `CMakeLists.txt` uses `add_subdirectory()` whenever `../supportlibs/librnp/CMakeLists.txt` exists, which is the case as soon as libretroshare is checked out inside the RetroShare super-project rather than standalone.

That inline build gets rnp's default backend, botan, and configure dies with

```
Could NOT find Botan (missing: BOTAN_LIBRARY BOTAN_INCLUDE_DIR)
(Required is at least version "2.14.0")
```

since the Android toolchain never builds botan.

Passing the same backend `build_librnp` uses fixes it. It is a no-op for standalone libretroshare checkouts, where `../supportlibs/librnp` does not exist and the pre-built rnp from the sysroot is picked up by `find_library()` instead — which is why CI never saw this.

Reported on IRC by defnax, who hit it building the AAR for rs-mobile from a super-project checkout.

### DIFF
--- a/misc/Android/prepare-toolchain-clang.sh
+++ b/misc/Android/prepare-toolchain-clang.sh
@@ -918,7 +918,14 @@ build_libretroshare()
 	rm -rf $B_dir
 	mkdir $B_dir || return $?
 	pushd $B_dir || return $?
+	## CRYPTO_BACKEND matters only when libretroshare CMake ends up building
+	## librnp inline, which happens when it is checked out inside the
+	## RetroShare super-project, as ../supportlibs/librnp then exists. Without
+	## it rnp defaults to botan, which the Android toolchain does not provide,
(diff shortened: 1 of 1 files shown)


jolavillette:
Superseded by #375. Pinning the backend removed the `Could NOT find Botan` error but only exposed the next one: the inline librnp build also has to *run* the `findopensslfeatures` probe it just compiled, which is an Android binary the build host cannot execute. The right fix is not to build librnp inline on Android at all, since the toolchain script already installed one in the sysroot.

PR #376, opened 2026-08-17
Android: link libdl into rnp's OpenSSL feature probe

jolavillette:
`build_librnp` fails while rnp configures itself:

```
ld.lld: error: undefined symbol: dlopen
>>> referenced by dso_dlfcn.c
>>>               dso_dlfcn.o:(dlfcn_load) in archive .../sysroot/usr/lib/libcrypto.a
```
```
CMake Error at cmake/Modules/FindOpenSSLFeatures.cmake:149 (message):
  Error building findopensslfeatures
```

rnp compiles a small `findopensslfeatures` helper and runs it to enumerate the
OpenSSL features. When cross-compiling it links that helper statically, so it
does not need the Android dynamic linker:

```cmake
if(CMAKE_CROSSCOMPILING_EMULATOR)
  target_link_options(findopensslfeatures PRIVATE -static)
endif(CMAKE_CROSSCOMPILING_EMULATOR)
```

but OpenSSL 1.1.1's `libcrypto.a` pulls in `dso_dlfcn.o`, which references the
`dl*` API, and CMake's `FindOpenSSL` does not add `${CMAKE_DL_LIBS}` to the
`OpenSSL::Crypto` interface here. The link line ends up as

```
clang -O3 -DNDEBUG -static ...o -o findopensslfeatures  .../libcrypto.a -pthread
```

with nothing providing `dlopen`, `dlsym`, `dlclose` and `dlerror`.

The NDK does ship a static `libdl.a` defining those four for real, so it is
enough to link it. It is added as a linked library rather than through
`CMAKE_EXE_LINKER_FLAGS`, because CMake places that variable *before* the
objects, where a static linker has no undefined symbol to resolve yet and drops
the archive silently.

This is arguably rnp's bug — their `-static` path looks untested against a
static OpenSSL 1.1.1 — so the patch stays a `sed` next to the existing ones, and
should be dropped once rnp handles it.

### Verified

`build_librnp` completes and installs `librnp.a` in the sysroot, and the whole
AAR now builds: 42 MB, `jni/arm64-v8a/libretroshare.so` exporting `JNI_OnLoad`,
the four `org.retroshare.service` classes in `classes.jar`. Ubuntu 24.04, NDK
`29.0.14206865`, API level 24, `arm64-v8a`, on top of #373 and #375.

### DIFF
--- a/misc/Android/prepare-toolchain-clang.sh
+++ b/misc/Android/prepare-toolchain-clang.sh
@@ -878,6 +878,23 @@ endif(CMAKE_CROSSCOMPILING_EMULATOR)' \
 	sed -i 's|COMMAND "${FOF}" "${feature}"|COMMAND ${FOF} "${feature}"|' \
 		"${S_dir}/cmake/Modules/FindOpenSSLFeatures.cmake"
 
+	# The feature probe is linked statically, and OpenSSL 1.1.1 libcrypto.a
+	# pulls in dso_dlfcn.o which references dlopen/dlsym/dlclose/dlerror.
+	# CMake's FindOpenSSL does not put ${CMAKE_DL_LIBS} in the OpenSSL::Crypto
+	# interface here, so nothing provides them and the link dies with
+	#
+	#   ld.lld: error: undefined symbol: dlopen
+	#   >>> referenced by dso_dlfcn.c
+	#   >>>               dso_dlfcn.o:(dlfcn_load) in archive .../libcrypto.a
+	#
+	# The NDK ships a static libdl.a defining those symbols, so link it
+	# explicitly. It has to come after libcrypto.a for the static linker to
+	# pick it up, hence a linked library rather than CMAKE_EXE_LINKER_FLAGS,
+	# which CMake places before the objects.
+	sed -i \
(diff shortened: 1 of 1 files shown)


G10h4ck:
@jolavillette  AFAIU an upstream patch should be submitted too, just like I did to fix librnp+openssl build on Android. Am I wrong? Can you deal with it?

jolavillette:
Not wrong — same file as your #2396. Upstream PR opened: rnpgp/rnp#2473. It links `${CMAKE_DL_LIBS}` inside the `CMAKE_CROSSCOMPILING_EMULATOR` static branch (no-op where it is empty). Root cause is that `FindOpenSSL` only propagates `CMAKE_DL_LIBS` to `OpenSSL::Crypto` when pkg-config lists it in `Libs.private`, which a bare cross sysroot does not have.

Verified with master's script (i.e. without the `sed` from this PR) on the existing 24/arm64 toolchain: rnp `main` (ee97ed76) still fails with the four `undefined symbol: dl*`, the patched branch configures, runs the probe under qemu and installs `librnp.a`.

Pushed a small follow-up commit here: the comment claimed the library had to come after `libcrypto.a`; that is true for GNU ld but the NDK's lld resolves it either way, so I reworded it and referenced the upstream PR so the `sed` can be dropped once it lands. The `sed` stays harmless after that (a duplicate `dl`).

PR #379, opened 2026-08-22
Contacts status

defnax:
Implemented GXS contact presence with getContactsStatus....

Included:
- away, busy, and online contact states
- Per-identity custom-state strings
- Contact-only exchange over authenticated GXS tunnels
- Persistent own status/custom state
- Transient remote-contact status
- CONTACT_STATUS_CHANGED identity events
- JSON API annotations

the GXS contact status is restricted per sender’s local contact list.
Current protections:
- Status is sent only when isARegularContact(contactId) returns true.
- Non-contact tunnels do not receive the new RS_CHAT_FLAG_GXS_CONTACT_STATUS packet.
- Incoming status is stored only if the authenticated sender is in the receiver’s regular contact list.
- Exchange occurs through the encrypted/authenticated GXS distant-chat tunnel.
- getContactsStatusList() returns entries only from mContacts.
- Each side decides independently: if Alice considers Bob a contact but Bob does not consider Alice one, Alice may advertise to Bob, but Bob will not advertise to Alice.

image


csoler:
Allowing a node to request contact status of GXS ids is problematic because it allows to make online/offline statistics that correlate a node with the anonymous ids it may host.

Obviously we could limit this status thing to signed IDs only, but then it's going to look unbalanced to have a status for some ids only. I don't really have any suggestion on how to solve this properly.

PR #380, opened 2026-08-22
I2P/SAM3: move the session restart delay off the stop path, it delayed shutdown by 10 s

jolavillette:
`p3I2pSam3::stopSession()` ends with an unconditional ten second sleep:

```cpp
// At least i2pd doesn't like to instantaniously stop and (re)start a session, wait here just a little bit.
// Not ideal but does the trick.
// (This happens when using the "restart" button in the settings.)
doSleep(std::chrono::seconds(10));
```

The reason behind it is real: `ServerPage::restartSam()` posts a `stop` ticket immediately followed by a `start` ticket on the same service-thread queue, and `startSession()` itself calls `stopSession()` when a session is already open. Without the delay, i2pd is asked to recreate a session it has just closed.

But a stop is not always followed by a start. `RsServer::rsGlobalShutDown()` calls `rsAutoProxyMonitor::stopAllRSShutdown()`, which posts that same `stop` ticket and then waits — up to 15 s, polling once per second — for the service to acknowledge it. The sleep therefore holds the whole application exit back by ten seconds, every time, on every node with SAM3 enabled:

```
D  rsAutoProxyMonitor::stopAllRSShutdown() waiting for auto proxy service(s) to shut down 0/15 (remaining: 1)
D  rsAutoProxyMonitor::stopAllRSShutdown() waiting for auto proxy service(s) to shut down 1/15 (remaining: 1)
...
D  rsAutoProxyMonitor::stopAllRSShutdown() waiting for auto proxy service(s) to shut down 10/15 (remaining: 0)
```

That is 10.0 s out of a 21.8 s shutdown, spent sleeping. The "stop" button in the settings pays it too, just as uselessly.

This keeps the delay but moves it to where it is actually needed. `stopSession()` records when the session was closed; `startSession()` waits out whatever is left of the ten seconds before creating a new one. The guarantee i2pd relies on is unchanged — a session is still never created less than ten seconds after the previous one was closed, whichever path led there — but a stop that is not followed by a start now returns immediately.

Two details worth noting:

- the wait sits *after* the `stopSession()` that `startSession()` may perform itself, so the restart path is covered as well as the two-ticket one;
- `mLastSessionClosed` starts at `time_point::min()` and the comparison is done on a deadline rather than a subtraction, so the first session of the process never waits and there is no overflow.

Measured on the same node after the change: the monitor logs a single `0/15 (remaining: 0)` line and the shutdown went from 21.8 s to 8.6 s.

Two files, no API change. Built with CMake and run for a couple of hours on Linux/Qt5 as part of a local integration branch — that is where the numbers come from — but not built standalone against master.

### DIFF
--- a/src/services/autoproxy/p3i2psam3.cpp
+++ b/src/services/autoproxy/p3i2psam3.cpp
@@ -29,6 +29,14 @@ static const std::string kConfigKeyBOBAddr     = "BOB_ADDR";
 
 static constexpr bool   kDefaultSAM3Enable = false;
 
+/* At least i2pd doesn't like to instantaniously stop and (re)start a session, so
+ * a new session is never created less than this long after the previous one was
+ * closed. The wait happens in startSession(), where it is actually needed, and
+ * not in stopSession(): a stop is not always followed by a start (RS shutdown,
+ * "stop" button in the settings) and waiting there held back the whole
+ * application exit by 10 seconds. */
+static constexpr auto kSessionRestartDelay = std::chrono::seconds(10);
+
 RS_SET_CONTEXT_DEBUG_LEVEL(2)
 
 static void inline doSleep(std::chrono::duration<long, std::ratio<1,1000>> timeToSleepMS) {
@@ -36,7 +44,9 @@ static void inline doSleep(std::chrono::duration<long, std::ratio<1,1000>> timeT
 }
 
 p3I2pSam3::p3I2pSam3(p3PeerMgr *peerMgr) :
-    mConfigLoaded(false), mPeerMgr(peerMgr), mPending(), mLock("p3i2p-sam3")
+    mConfigLoaded(false), mPeerMgr(peerMgr), mPending(),
+    mLastSessionClosed(std::chrono::steady_clock::time_point::min()),
+    mLock("p3i2p-sam3")
 #ifdef RS_USE_I2P_SAM3_LIBSAM3
   , mLockSam3Access("p3i2p-sam3-access")
 #endif
@@ -498,6 +508,19 @@ bool p3I2pSam3::startSession()
 		stopSession();
 	}
 
+	// Give i2pd time to let go of the previous session. No lock may be held here.
+	std::chrono::steady_clock::time_point sessionReady;
+	{
+		RS_STACK_MUTEX(mLock);
+		sessionReady = mLastSessionClosed + kSessionRestartDelay;
+	}
+	const auto now = std::chrono::steady_clock::now();
+	if (now < sessionReady) {
+		const auto wait = std::chrono::duration_cast<std::chrono::milliseconds>(sessionReady - now);
+		RS_DBG("waiting ", wait.count(), " ms before creating a new SAM session");
+		doSleep(wait);
+	}
+
 	auto session = (Sam3Session*)rs_malloc(sizeof (Sam3Session));
 
 	// add nick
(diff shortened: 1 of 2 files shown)


csoler:
Wouldn't it make sense to fix the problem of queued events in i2pd (assuming it is part of RS code) ?

jolavillette:
The two queued tickets (`stop` then `start`, posted by `ServerPage::restartSam()`) are indeed RS code, but they are not what needs the delay. A single `start` has the same problem, because `startSession()` calls `stopSession()` itself when a session is already open. Whatever the path, the sequence is always "close the SAM session, then create a new one with the same private key", and that is what i2pd does not like.

As far as I can tell from reading i2pd (2.61.0 and current master), the constraint is on their side, not in RS. SAM v3 has no "session remove" command for a single session: a session is closed by closing its control socket, which is all `sam3CloseSession()` does. i2pd then tears the session down asynchronously: `SAMBridge::CloseSession()` stops the streams and, in the default single-thread mode, postpones the session cleanup with a 5-second timer; only when the session object finally dies does `ClientContext::DeleteLocalDestination()` remove the destination (keyed by identity hash) and stop it. If a new `SESSION CREATE` with the same key arrives inside that window, `ClientContext::CreateNewLocalDestination()` finds the existing destination, logs "Local destination ... exists" and hands it to the new session, and the reply is `RESULT=OK`. I have not traced it with a live router, but from the code the pending cleanup of the old session then erases and stops the destination the new session is using — `DeleteLocalDestination()` does not check the session refcount.

If that reading is right, there is nothing to retry on and nothing RS can query; the only observable is time. That is why sehraf put a fixed delay there in 2020, and I kept it. The proper fix would be in i2pd (do not delete a destination that has been re-acquired by a newer session); I can open an issue at PurpleI2P/i2pd to confirm it with them. Either way RS has to work with the routers people run today.

What this PR changes is only *where* the delay is paid: on the `start` that follows a `stop`, instead of on every `stop`. Shutdown and the "stop" button never create a new session, so they no longer wait 10 seconds for nothing; the restart path keeps exactly the same guarantee as before.

jolavillette:
Raised upstream: PurpleI2P/i2pd#2558 (issue) and PurpleI2P/i2pd#2559 (fix). Once that lands in a released i2pd, the delay in startSession() can go.

jolavillette:
Update on the i2pd side, since that was the open question here.

I raised it upstream (PurpleI2P/i2pd#2558). orignal confirmed the bug is real and, rather than take a patch, wrote his own fix in the `openssl` (development) branch: commit `86cbb4b0`. It is in no release yet; the latest release, 2.61.0, still has the old behaviour.

What his fix changes: when a session is closed and a new one is created with the same keys while the old destination is still being torn down, the router now answers `SESSION STATUS RESULT=DUPLICATED_DEST` for a few seconds instead of accepting the session and then silently killing its destination. So the failure becomes visible and recoverable, but the contract changes: the client must retry on `DUPLICATED_DEST` instead of assuming the session is up.

I verified `86cbb4b0`: it covers exactly the RS case (close, then re-create with the same key). Two related cases are still broken (a second session created before the first is closed, and a tunnel from `tunnels.conf` sharing keys with a SAM session), because the session destructor still deletes the destination unconditionally; the general fix will likely move to `ClientContext`. That is upstream's call now.

For this PR nothing changes today: no released i2pd has any of this, so the delay is still required, and moving it off shutdown/stop is the right step. Once a fixed i2pd ships, the follow-up on our side is to replace the sleep with a retry on `DUPLICATED_DEST` in `startSession()`; I can do that in a separate PR when the time comes.

PR #383, opened 2026-08-30
Distant chat: tell every client when a conversation is closed locally

jolavillette:
Stacked on #378 (`fix/distant-chat-close-keeps-contact`): the first commit here is that PR's, this one adds the second. Merge #378 first and this reduces to one commit.

`closeDistantChatConnexion()` posted no event. The client that called it knew; the others did not: a conversation opened and then left from the web UI stayed open — green LED and all — in the desktop window the core had popped for it, and closing that window then offered to end a conversation that no longer existed. Polling cannot fix that: once the tunnel is gone, `getDistantChatStatus()` answers the same false for a conversation the peer closed (where the window should stay open) and for one closed from another client (where it should go); only an explicit event tells the two apart.

This adds `RsDistantChatEventCode::TUNNEL_STATUS_LOCALLY_CLOSED` (0x05) and posts it from `closeDistantChatConnexion()` whenever something was actually closed, with the tunnel id, so every client of this core — desktop GUI, web UI, JSON API users — drops the conversation at the same time. The remote close path is untouched: `markDistantChatAsClosed()` stays as it was, and `REMOTELY_CLOSED` keeps coming from `notifyTunnelStatus()`.

GUI counterpart: RetroShare/RetroShare#3297 (closes the popup on this event).

### DIFF
--- a/src/chat/distantchat.cc
+++ b/src/chat/distantchat.cc
@@ -360,7 +360,25 @@ bool DistantChatService::closeDistantChatConnexion(const DistantChatPeerId &tunn
     // the web UI, any JSON API caller -- report a conversation as closed when
     // nothing had been closed at all, the tunnel having died on its own before.
 
-    return tunnel_closed || contact_removed ;
+    bool closed = tunnel_closed || contact_removed ;
+
+    // The client that asked for the close knows about it; the others do not,
+    // and a conversation closed from the web UI stayed open in the desktop
+    // chat window. Polling cannot fix that: once the tunnel is gone,
+    // getDistantChatStatus() answers the same false for a conversation the
+    // peer closed (where the window should stay open) and for one we closed
+    // from another client (where it should go). Only an explicit event tells
+    // them apart, so post it here -- and every client of this core drops the
+    // conversation at the same time.
+    if(closed && rsEvents)
+    {
+        auto ev = std::make_shared<RsDistantChatEvent>();
+        ev->mEventCode = RsDistantChatEventCode::TUNNEL_STATUS_LOCALLY_CLOSED;
+        ev->mId = tunnel_id;
+        rsEvents->postEvent(ev);
+    }
+
+    return closed ;
 }
 
 uint32_t DistantChatService::getDistantChatPermissionFlags()
--- a/src/chat/distantchat.cc
+++ b/src/chat/distantchat.cc
@@ -327,11 +327,40 @@ bool DistantChatService::getDistantChatStatus(const DistantChatPeerId& tunnel_id
 
 bool DistantChatService::closeDistantChatConnexion(const DistantChatPeerId &tunnel_id)
 {
-    mGxsTunnels->closeExistingTunnel(RsGxsTunnelId(tunnel_id), DISTANT_CHAT_GXS_TUNNEL_SERVICE_ID) ;
-    
-    // also remove contact. Or do we wait for the notification?
-    
-    return true ;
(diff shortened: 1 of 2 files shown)


src/chat/distantchat.cc
```diff
@@ -327,11 +320,49 @@ bool DistantChatService::getDistantChatStatus(const DistantChatPeerId& tunnel_id
...
+            mDistantChatContacts.erase(it) ;
+            contact_removed = true ;
+        }
+    }
+
+    // Answering true whatever happened made every client -- the chat window,
+    // the web UI, any JSON API caller -- report a conversation as closed when
+    // nothing had been closed at all, the tunnel having died on its own before.
+
+    bool closed = tunnel_closed || contact_removed ;
```
csoler, line 363: this test is redundant: It's not possible that a distant chat contact is not in the list while a tunnel exist for that contact. The only relevant info here is tunnel_closed. What you should do here is test that tunnel_closed is equal to contact_removed. If not, display an error. Another way to do it is to return false early if closeExistingTunnel() returns false, and then (meaning it returned true) display an error is mDistantChatContacts does not contain the tunnel_id. Don't account for the error and send the event anyway after that.

csoler (changes requested):
The solution presented here does not seem appropriate to me because eventually both methods (closeDistantChatConnexion and markDistantChatAsClosed will generate TUNNEL_STATUS_LOCALLY_CLOSED, which is inconsistent).

What you should do is
- keep these two functions follow separate paths
- no need to modify markDistantChatAsClosed() to send the event REMOTELY_CLOSED because this is already done in DistantChatService::notifyTunnelStatus() (called by turtle when the tunnel is remotely closed)
- in closeDistantChatConnexion(), send the event LOCALLY_CLOSED and remove the contact from mDistantChatContacts (was that forgotten??)
- update the UI accordingly

Note that we cannot rely on turtle for the "locally closed" event, because p3GxsTunnel ONLY calls p3turtle::stopMonitoringTunnels() when the tunnel is a client side tunnel.

jolavillette:
Reworked as suggested, thanks. `markDistantChatAsClosed()` is back to exactly what it was — the two paths stay separate, and `REMOTELY_CLOSED` keeps coming from `notifyTunnelStatus()` alone. `closeDistantChatConnexion()` now removes the contact and posts `LOCALLY_CLOSED`; the GUI counterpart (RetroShare/RetroShare#3297) closes the chat window on that event only, so a remote close still leaves the window open as before.

On the inline comment: I kept the contact-removal leg instead of returning false early when `closeExistingTunnel()` fails, because a still-registered contact whose tunnel entry is already gone is a state that actually occurs — it is where the "(EE) Cannot close distant tunnel connection" log in the description came from:

* when the peer closes the conversation, `handleRecvStatusItem()` can mark our tunnel entry `REMOTELY_CLOSED` (depending on the tunnel direction), and `flush()` erases such entries 20s later (p3gxstunnel.cc:205) — from then on `closeExistingTunnel()` answers false for that tunnel id;
* nothing removes the DistantChatService contact in the meantime: its only trigger is the `RS_CHAT_FLAG_CLOSING_DISTANT_CONNECTION` chat item (distantchat.cc:128), and nothing sends that item any more — its last sender went away in c521e4ed1 (2015), with the move to GXS tunnels.

So closing such a conversation, from any client, can find no tunnel left but a live contact. Returning false early there would leave the contact behind and post no event — the original bug. The converse (a closed tunnel with no registered contact) is a genuine anomaly and now logs an (EE) error, as you suggested.

PR #386, opened 2026-09-08
build(cmake): add RS_SYSTEM_LIBRNP to link the system librnp

jolavillette:
New option **`RS_SYSTEM_LIBRNP`** (default **OFF**, requires `RS_RNPLIB`): link the distribution's librnp instead of building the vendored copy. Required by Debian policy (no embedded copies of libraries the distribution already packages), and useful for any distro packaging.

It reuses the pre-built-librnp code path introduced for Android:

- a missing system librnp is a **fatal configure error** — no silent fallback to the vendored sources, which would defeat the point for distribution builds;
- the headers are resolved via `find_path` so non-standard prefixes work, and exported **PUBLIC**: `rnppgphandler.h` (reached from the public `authgpg.h`) includes `rnp/rnp.h`, so consumers of the `retroshare` target need them too — the vendored code path exports its include directory PUBLIC for the same reason.

Note: current master requires **rnp >= 0.17.1** (`rnp_key_direct_signature_create` / `rnp_key_signature_sign`, absent from 0.17.0). Validated by a full Debian package build against Debian sid's rnp 0.18.1 in a clean container.

No default behavior changes: with the option OFF (the default) the vendored build is untouched.

Companion PR on the super-project side: RetroShare/RetroShare#3309. Part of the groundwork for Debian packaging (Debian bug #659069).


G10h4ck:
The PR looks fine, but comments really need to be that verbose? Use succinct comments when necessary and explain why and do not restate what the code does

jolavillette:
Fair point, trimmed: the option description is now one line and the two comments only say why (distro policy forbids the vendored copy, hence the hard failure; PUBLIC because rnp/rnp.h is reached from authgpg.h). Code unchanged, force-pushed.

PR #390, opened 2026-09-13
Turtle: expose rsTurtle router info and traffic statistics to JSON API

defnax:
image


src/retroshare/rsturtle.h
```diff
@@ -110,12 +156,32 @@ class RsTurtle
...
 		virtual bool enabled() const = 0 ;
 
-		// This is temporary, used by Operating Mode.
-		virtual void setSessionEnabled(bool) = 0 ;
+		/**
+		 * @brief Enable or disable turtle router for current session
+		 * @jsonapi{development}
+		 * @param[in] enable pass true to enable, false to disable
+		 */
+		virtual void setSessionEnabled(bool enable) = 0 ;
```
csoler: I makes no sense to disable turtle since it is used almost everywhere: distant chat, FT, distant GXS sync, email, etc. Allowing the user to control this has complicated consequences => non userfriendly UI

src/retroshare/rsturtle.h
```diff
@@ -150,22 +216,61 @@ class RsTurtle
...
 		virtual void getTrafficStatistics(TurtleTrafficStatisticsInfo& info) const = 0;
 
-		// Convenience function.
+		/**
+		 * @brief Check if peer is turtle peer
+		 * @jsonapi{development}
+		 * @param[in] peer_id peer id to check
+		 * @return true if turtle peer, false otherwise
+		 */
 		virtual bool isTurtlePeer(const RsPeerId& peer_id) const = 0 ;
```
csoler, line 228: same here: this function is actually used to determine is a virtual peer name is handled by p3turtle or not. This is really not useful in the UI and may bring confusion.

csoler (changes requested):
Some functions shouldn't be exposed (see comments). Otherwise the rest of it is ok.

PR #391, opened 2026-09-19
ft, util: harden directory creation around FileRequest()/moveFile()

SchlauFuchs:
Companion fixes for RetroShare/RetroShare#3312 (which removes GUI-side `mkpath()` calls on the assumption that the transfer engine already creates any destination directory it needs), per @jolavillette's review there:

- `ftController::FileRequest()`'s `size==0` stub-file path writes directly into `dest/fname` without creating `dest` first. Unlike a normal download, this file is never routed through `moveFile()` at completion (there is no transfer to complete), so nothing else creates that directory - a zero-byte file inside a not-yet-created sub-folder of a directory or collection download was silently lost. Create the destination directory first, the same way `moveFile()` does for every other file.

- `RsDirUtil::moveFile()` called the throwing single-argument overload of `std::filesystem::create_directories()` with nothing above it to catch an exception (it runs on the file transfer thread). A destination name that is merely invalid on the current filesystem (e.g. a character disallowed on Windows) would abort the whole process here, and again on every subsequent restart, since the same completion is retried. Use the non-throwing `error_code` overload instead, matching the existing `RsErr()`-and-`return false` error handling right below it.

This is independent of #388 (the `size==0`/empty-hash refusal fix) - based directly on current `master`, no dependency between the two.

## Testing

Clean rebuild of `libretroshare.a`, no new warnings.

https://claude.ai/code/session_0137V8RpdwiKisg9SSrMEiBt


SchlauFuchs:
Closing this - on reflection this fix isn't actually independent of #388 (both touch the identical `if(size==0)` block), so I've folded it into that PR as a second commit instead of maintaining two overlapping branches.

